PR #5, opened 2016-06-21
Feature/esp slip communication

### LINKED ISSUE #2: WiFi upload should work with platformio esptool
The version of the esp-tool embedded in platformio does not work with our `program-esp` firmware. This is probably a timing issue, but needs to be debugged.

The work-around is to use another esp programmer, as demonstrated in the [Makefile](https://github.com/sarfata/kbox-firmware/blob/master/Makefile#L13) but that only works on Linux and Mac OS.

sarfata:
This fixes #2 and makes it easier to upload firmware via standard platformio commands.

This also improves the "programmer" code by reading in frames instead of reading in bytes.


sarfata:
Successfully tested on Linux/Windows/Mac.

PR #8, opened 2016-09-07
Feature/update nmea2000

sarfata:


sarfata:
Update NMEA2000 and store pressure in Pascal to follow ttlappalainen/NMEA2000#18

PR #9, opened 2016-09-08
Feature/revisit kmessage

sarfata:


sarfata:
Tested with NMEA2000 setup at home. All look good.

PR #25, opened 2016-10-17
Signalk embedded

sarfata:
(work in progress)


sarfata:
Abandonning this PR. 

#31 is more advanced and a better solution with less dynamic memory allocation and a C++ style interface.

PR #28, opened 2017-03-21
Rigm edits

matrixlinks:
Change to enable configuring Kbox orientation and autopilot functionality


lib/KBox/src/tasks/AutoPilotTask.cpp
```diff
@@ -0,0 +1,136 @@
...
+        apCurrentHeading = nav.getCurrentHeading();
+        apTargetHeading = nav.getTargetHeading();
+        apRudderSensorPosition = nav.getRudderSensorPosition();
+    
+       // DEBUG("apPilotCurrentHeading - %.0f",apCurrentHeading);
+}
+
+void AutoPilotTask::loop() {
+    
+    if (isSetup == false){ //tried a separate "setup" for init but could not get it to work
```
sarfata: Please try moving this back into init and I will help you debug it if it does not work.

lib/KBox/src/tasks/AutoPilotTask.cpp
```diff
@@ -0,0 +1,136 @@
...
+    headingPID.SetSampleTime(250);
+    DEBUG("Init autopilot complete");
+    isSetup = true;
+    }
+    
+    headingPID.Compute();
+    
+    autoPilotOffCourse = apTargetHeading - apCurrentHeading; //variable not currently used for anything
+    
+    autoPilotDeadZone = 5; //pending tuning on water
```
sarfata: We should put all the tuning parameters in one place, ideally as `#define` or `static const` variables in the header file. This way it's easy to see what values can be tuned.

lib/KBox/src/tasks/IMUTask.cpp
```diff
@@ -24,6 +24,10 @@
 #include "IMUTask.h"
 #include "KBoxDebug.h"
 
+const String kBoxOrientation = "bow"; //Modification to send correct course, pitch, roll data based on where the back of the KBOX is mounted.  Options are back of the box to port, to bow, to stbd, to stern
```
sarfata: Can you make that an enum instead of a string?

lib/KBox/src/tasks/IMUTask.cpp
```diff
@@ -47,8 +51,29 @@ void IMUTask::loop() {
     eulerAngles = bno055.getVector(Adafruit_BNO055::VECTOR_EULER);
     //DEBUG("Vector Euler x=%f y=%f z=%f", eulerAngles.x(), eulerAngles.y(), eulerAngles.z());
     //DEBUG("Course: %.0f MAG  Pitch: %.1f  Heel: %.1f", eulerAngles.x(), eulerAngles.y(), eulerAngles.z());
-    IMUMessage m(sysCalib, eulerAngles.x(), /* yaw?*/ 0, eulerAngles.y(), eulerAngles.z());
-    sendMessage(m);
+    //IMUMessage m(sysCalib, eulerAngles.x(), /* yaw?*/ 0, eulerAngles.y(), eulerAngles.z()); //original
+    //sendMessage(m);
```
sarfata: I think things would be a bit easier to read if you first assigned values to double variables for yaw, heel and pitch and then applied the transformation, and at the end built the message and sent it.

lib/KBox/src/tasks/NMEA2000Task.cpp
```diff
@@ -113,6 +113,8 @@ void NMEA2000Task::visit(const IMUMessage &m) {
     sendN2kMessage(n2kmessage);
 
     _imuSequence++;
+  } else {
+       DEBUG("IMU is not calibrated: move the unit around to calibrate it"); //debug message added to warn user if IMU not calibrated
```
sarfata: That is going to be a lot of debugging messages, no? Maybe we should remove this one and instead add some UI to show the user whether the IMU is calibrated.

lib/KBox/src/tasks/AutoPilotTask.h
```diff
@@ -0,0 +1,55 @@
...
+#define D_Param  0.0 //derivative or D-gain, the preact determines how much to change the OP due from a change in direction of the error
+
+//extern double autoPilotRudderCommand;
+
+class AutoPilotTask : public Task, public KReceiver, public KVisitor, public KGenerator {  private:
+    
+    
+    double apRudderSensorPosition;
+    double apRudderCommandSent;
+    double autoPilotOffCourse; //not currently used
```
sarfata: Please remove things that are not currently used. It will make it a lot easier to read and understand your code :)

lib/KBox/src/tasks/AutoPilotTask.cpp
sarfata: Also, I think we should do all variable initializations in `setup()`

lib/KBox/src/KBox.h
```diff
@@ -43,6 +43,7 @@
 #include "tasks/IMUTask.h"
 #include "tasks/NMEAReaderTask.h"
 #include "tasks/WiFiTask.h"
+#include "tasks/AutoPilotTask.h" //RIGM added
```
matrixlinks: Will remove.

lib/KBox/src/KMessage.h
```diff
@@ -145,41 +151,46 @@ class NMEA2000Message: public KMessage {
 
 class IMUMessage: public KMessage {
   private:
+    String label; //RIGM added, may not need it
```
matrixlinks: Will remove

lib/KBox/src/KMessage.h
```diff
@@ -145,41 +151,46 @@ class NMEA2000Message: public KMessage {
...
+        return label;
+    };
 
     int getCalibration() const {
       return calibration;
     };
 
     /*
-     * Heading in Radians
+     * Heading in Radians: for necessary precision need to use float, not double for radians (RIGM)
```
matrixlinks: I misunderstood the precise, will change back to double

lib/KBox/src/KMessage.h
```diff
@@ -191,6 +202,92 @@ class IMUMessage: public KMessage {
     };
 };
 
+class NAVMessage: public KMessage { //RIGM added for autopilot functionality
+private:
+    String label; //not sure we need a label
```
matrixlinks: Will remove

lib/KBox/src/KMessage.h
```diff
@@ -191,6 +202,92 @@ class IMUMessage: public KMessage {
...
+    bool getApWaypointMode() const {
+        return apWaypointMode;
+    };
+    
+    bool getApDodgeMode() const {
+        return apDodgeMode;
+    };
+    
+    /*
+     * Courses in Degrees, therefore using doubles:
```
matrixlinks: Will note that headings are magnetic.
With respect to the data passed back and forth between the nav page and the autopilot task, yes we could have the Kmessage heading-related variables in radians, but assuming we are doing the autopilot math in degrees, that would mean converting degrees to radians in the navPage and then immediately converting degrees to radians in the autopilot task.  Seems that this would add calculation overhead for no benefit.  On the other hand, the IMU Message is in radians, and this makes it easier to transmit directly to N2K.

lib/KBox/src/KMessage.h
```diff
@@ -191,6 +202,92 @@ class IMUMessage: public KMessage {
...
+    double getRudderSensorPosition() const {
+        return rudderSensorPosition;
+    };
+    
+    
+};
+
+class APMessage: public KMessage {  //RIGM added for autopilot functionality
+private:
+    String label; //not sure we need a label
```
matrixlinks: Will remove

lib/KBox/src/KMessageNMEAVisitor.cpp
```diff
@@ -25,6 +25,12 @@
 #include "KMessageNMEAVisitor.h"
 #include "util/NMEASentenceBuilder.h"
 #include "util/nmea2000.h"
+//RIGM added
+#include "tasks/NMEA2000Task.h"
+#include "KBoxDebug.h"
+
+#define IMU_IN_NMEA2K true;  //variable to determine whether IMU data sent out over N2K
```
matrixlinks: OK

lib/KBox/src/KMessageNMEAVisitor.cpp
```diff
@@ -63,7 +69,40 @@ void KMessageNMEAVisitor::visit(const NMEA2000Message &n2km) {
...
+    free(s);
+    
+    
+    s = nmea_pcdin_sentence_for_n2kmsg(n2kmessage2, millis() / 1000);
+    //DEBUG("nmeaContent: %s", s);
+    nmeaContent += String(s) + "\r\n";
+    free(s);
+    
+    
+#endif
```
matrixlinks: OK

lib/KBox/src/pages/NavigationPage.cpp
```diff
@@ -0,0 +1,305 @@
...
+#include <stdio.h>
+#include "KBoxDebug.h"
+
+
+bool apMode = false;
+bool apDodgeMode = false;
+bool apHeadingMode = false;
+bool apWaypointMode = false;//reserved for future use
+String apModeString = "APMode: Off"; // Modes: "Off", "Heading" == follow set heading, "Waypoint" = GoTo provided Waypoint, "Dodge"
+String apWaypointString = "045 20.2N, 081 50.3W";//placeholder for future use
```
matrixlinks: Do you mean the apWayMode boolean (which we will need later), or the apWaypointString.  I put data in the string just for kicks, but could take the string out completely or just initialize it to "".

lib/KBox/src/pages/NavigationPage.cpp
```diff
@@ -0,0 +1,305 @@
...
+        //DEBUG("navEncoderClicks - %d", navEncoderClicks);
+    return true;
+}
+
+
+String NavigationPage::formatMeasurement(float measure, const char *unit) {
+  // extra spaces at the end needed to clear previous value completely
+  // (we use a non-fixed width font)
+  char s[10];
+  //snprintf(s, sizeof(s), "%.1f %s  ", measure, unit);
```
matrixlinks: Sure.  That was your line, but I'd be happy to remove it for you if you want!

lib/KBox/src/pages/NavigationPage.cpp
```diff
@@ -0,0 +1,305 @@
...
+    if (vm.getLabel() == "bat3") {
+        rawRudderAngle = vm.getVoltage();
+        //provision for tuning
+        navRudderSensorPosition = rawRudderAngle;
+    }
+}
+
+
+void NavigationPage::visit(const APMessage &ap) { //this function never called
+
```
matrixlinks: OK

lib/KBox/src/tasks/ADCTask.cpp
```diff
@@ -33,11 +33,18 @@ void ADCTask::loop() {
   int bat1_adc = adc.analogRead(bat1_analog, ADC_0);
   int bat2_adc = adc.analogRead(bat2_analog, ADC_0);
   int bat3_adc = adc.analogRead(bat3_analog, ADC_0);
+    
+  //permits rudder sensor on bat3 input.  Eventually should move rudder sensor calcs to a separate task
```
matrixlinks: I was just taking a shortcut here for now.  I assume the better solution is to create a separate RudderSensor task and move the relevant code into that.

lib/KBox/src/tasks/AutoPilotTask.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+
+
+/*pending tuning,following is set in AutoPilotTask.h
+#define P_Param  0.5 //proportional param P-gain, the gain determines how much change the OP will make due to a change in error
+#define I_Param  0.0 //integral or I-gain, the the reset determines how much to change the OP over time due to the error
+#define D_Param  0.0 //derivative or D-gain, the preact determines how much to change the OP due from a change in direction of the error
+ */
+
+PID headingPID(&apCurrentHeading, &apTargetRudderPosition, &apTargetHeading, (double)P_Param, (double)I_Param, (double)D_Param, (int)REVERSE);
```
matrixlinks: I tore my hair out trying to figure out how to do that in c++, and resorted to the global variable when I couldn't figure out how to declare it in Autopilot.h file and access the object in the AutoPilot.cpp.  Can you show me the syntax for this?

lib/KBox/src/tasks/AutoPilotTask.cpp
```diff
@@ -0,0 +1,136 @@
...
+bool isSetup = false; //if false sets initial PID variables on bootup
```
matrixlinks: Won't need this if we have a setup function.  I did it this way for the same reason I set up headPID as a global .

lib/KBox/src/tasks/AutoPilotTask.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+void AutoPilotTask::processMessage(const KMessage &message) {
+    message.accept(*this);
+}
+
+
+void AutoPilotTask::visit(const NAVMessage &nav) {  //this function now working!!
+    
+        apCurrentHeading = nav.getCurrentHeading();
```
matrixlinks: OK

lib/KBox/src/tasks/AutoPilotTask.cpp
matrixlinks: Init or setup will work once I can access headingPID as a non-global.

lib/KBox/src/tasks/AutoPilotTask.cpp
matrixlinks: OK will move tuning variables as #defines and initialize in setup

lib/KBox/src/tasks/IMUTask.cpp
matrixlinks: Sure.  I'd do some research on enum ...

lib/KBox/src/tasks/IMUTask.cpp
matrixlinks: OK, I think I understand what you are getting at

lib/KBox/src/tasks/NMEA2000Task.cpp
matrixlinks: We could show a message like "Calibrating" in place of "APMode: Off" on the nav page until IMU is calibrated.  Could also prevent trying to use the AP until calibration is OK

lib/KBox/src/tasks/AutoPilotTask.h
matrixlinks: OK, but I've now just realized why that is there:  we will actually be using it.  But I get the general point.

A more general process newbie question:  once I've made these changes in my source code, what happens next.  Do I re-upload the respective files to the branch?  Do I attach the respective files to the pull request?  Do I generate another pull request based on up-loading the files?

lib/KBox/src/pages/NavigationPage.cpp
```diff
@@ -0,0 +1,305 @@
...
+ an object or to manually put it back on course.  Waypoint functionality will be added at a later time. RIGM
+ */
+
+#include "NavigationPage.h"
+#include "MFD.h"
+#include <stdio.h>
+#include "KBoxDebug.h"
+
+
+bool apMode = false;
```
matrixlinks: This is where my knowledge of c++ falls short.  How would I make these variables a member of the class?

With respect to having them as an enum, that would present some challenges.  I set them up as separate booleans because because they are not all mutually exclusive.  If we are in heading mode and switch to dodge mode, we need to know to go back to heading mode when we come out of dodge mode, rather than Waypoint mode.  Heading and Waypoint are mutually exclusive, but Dodge is not.  For now I suggest we stick with the booleans.

lib/KBox/src/KBox.h
matrixlinks: now removed throughout

lib/KBox/src/KMessage.h
matrixlinks: now removed

lib/KBox/src/KMessage.h
matrixlinks: reverted to doubles

lib/KBox/src/KMessage.h
matrixlinks: now removed

lib/KBox/src/KMessage.h
matrixlinks: Pending discussion about whether to do conversions on messages passed back and forth between navPage and autopilot task.

lib/KBox/src/KMessage.h
matrixlinks: now removed

lib/KBox/src/KMessageNMEAVisitor.cpp
matrixlinks: fixed

lib/KBox/src/KMessageNMEAVisitor.cpp
matrixlinks: fixed

lib/KBox/src/pages/NavigationPage.cpp
matrixlinks: now left as blank string

lib/KBox/src/pages/NavigationPage.cpp
matrixlinks: removed, I now realize I did change it to get rid of the decimals on course headings

lib/KBox/src/pages/NavigationPage.cpp
matrixlinks: done

lib/KBox/src/tasks/ADCTask.cpp
matrixlinks: have created new RudderSensor task

lib/KBox/src/tasks/AutoPilotTask.cpp
matrixlinks: fixed

lib/KBox/src/tasks/AutoPilotTask.cpp
```diff
@@ -0,0 +1,136 @@
...
+        apTargetHeading = nav.getTargetHeading();
+        apRudderSensorPosition = nav.getRudderSensorPosition();
+    
+       // DEBUG("apPilotCurrentHeading - %.0f",apCurrentHeading);
+}
+
+void AutoPilotTask::loop() {
+
```
matrixlinks: fixed

lib/KBox/src/tasks/AutoPilotTask.cpp
matrixlinks: now set up as defines

lib/KBox/src/tasks/IMUTask.cpp
matrixlinks: now set up as enum

lib/KBox/src/tasks/IMUTask.cpp
matrixlinks: done

lib/KBox/src/tasks/NMEA2000Task.cpp
matrixlinks: navpage now shows "calibrating" message until imu is calibrated

lib/KBox/src/tasks/AutoPilotTask.h
matrixlinks: In th course of making these changes I implemented quite a few improvements in how the ap code works.  So now wondering if I should upload modified and corrected files and generate a new pull request.

sarfata:
The beauty of a code request is that it can be updated :)

Just commit your new changes to your branch and click the sync button, this
page will update with your updated code.

matrixlinks:
Hi Thomas:

I have uploaded all the changes made so far.  However, so far have not been able to find a “sync” button, but I see that there are references in the pull request to my recent uploads.  If there is some other button I should be clicking, let me know.

The two open issues are:

• Whether to do DegToRad / RadToDeg on all messages exchanged between the NavPage and the AutoPilot Task: since these are not going out to N2K, it seems to me that doing conversions on all those messages could be adding unnecessary calculation overhead: but you have a better idea than I of the implications of that.  

• I need some help with the syntax for declaring the PID object in AutoPilotTask.h so that it is accessible in AutoPilotTask.cpp without being declared it as a global variable.  I tried various formats without success: this is where my understanding of c++-style code falls short.

Other than those two issues, I think everything else is cleaned up.

A bientot,

Rob

lib/KBox/src/pages/NavigationPage.cpp
```diff
@@ -0,0 +1,280 @@
...
+            apModeDisplay->setText(apModeString);
+            apModeDisplay->setColor(ColorGreen);
+        }
+    } else {
+        apModeString = "Calibrating   ";
+        apModeDisplay->setText(apModeString);
+        apModeDisplay->setColor(ColorRed);
+    }
+    
+    rawHeading = RadToDeg(imu.getCourse());
```
sarfata: I think it would make sense to fix the IMU to always return something between 0 and 360. A heading of -30 does not make a lot of sense I think.

lib/KBox/src/KMessageNMEAVisitor.cpp
```diff
@@ -25,6 +25,12 @@
 #include "KMessageNMEAVisitor.h"
 #include "util/NMEASentenceBuilder.h"
 #include "util/nmea2000.h"
+//RIGM added
+#include "tasks/NMEA2000Task.h"
+#include "KBoxDebug.h"
+
+#define IMU_IN_NMEA2K true  //variable to determine whether IMU data sent out over N2K
+unsigned int _imuSequence = 1;
```
sarfata: This is a bit of a hack and we could hide it better by making this variable a static variable in the function where it is used. We do not need anywhere else.

lib/KBox/src/KMessage.h
```diff
@@ -191,6 +198,97 @@ class IMUMessage: public KMessage {
...
+        return targetRudderPosition;
+    };
+    
+    double getRudderCommandSent() const {
+        return rudderCommandSent;
+    };
+    
+};
+
+class RUDMessage: public KMessage {  //RIGM added for autopilot functionality
```
sarfata: Please remove the `//RIGM`, here and above.

lib/KBox/src/KMessage.h
```diff
@@ -157,20 +164,20 @@ class IMUMessage: public KMessage {
     void accept(KVisitor &v) const {
       v.visit(*this);
     };
-
+
```
sarfata: I am really being nitpicky here but please remove those extra spaces here.

lib/KBox/src/tasks/ADCTask.cpp
sarfata: I think an even better solution would be to keep ADCTask and have a config option in it to replace BAT3 by a Rudder sensor.

If the option is not set, ADCTask would send four VoltageMeasurement like it used to.
If the option is set, ADCTask would send three VoltageMeasurement and one RudderAngle.

lib/KBox/src/tasks/AutoPilotTask.cpp
```diff
@@ -0,0 +1,135 @@
...
+#include "../drivers/board.h"
+#include "KMessage.h"
+#include <NMEA2000.h>
+#include <N2kMessages.h>
+
+double apCurrentHeading = 0;
+double apTargetHeading = 0;
+double apTargetRudderPosition = 0;
+
+PID headingPID(&apCurrentHeading, &apTargetRudderPosition, &apTargetHeading, (double)P_Param, (double)I_Param, (double)D_Param, (int)REVERSE);
```
sarfata: What you need here is to declare the headingPID and the current and target variables in the class definition:
```c++
class AutoPilotTask {
  private:
    double apCurrentHeading, apTargetHeading, apTargetRudderPosition;
    PID headingPID;
  // ...
};
```

And initialize it in the constructor:
```c++
AutoPilotTask::AutoPilotTask() : apCurrentHeading(0), apTargetHeading(0), apTargetRudderPosition(0), headingPID(&apCurrentHeading, &apTargetRudderPostion, &apTargetHeading, (double)P_Param, (double)I_Param, (double)D_Param, (int) REVERSE)
{
}
```

lib/KBox/src/tasks/AutoPilotTask.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+void AutoPilotTask::visit(const RUDMessage &rm) {
+    apRudderSensorPosition = rm.getRawRudderAngle();
+}
+
+void AutoPilotTask::loop() {
+    
+  if (isSetup == false){ //tried a separate "setup" for init but could not get it to work
+    sameLastDirection=true;
+    apTargetRudderPosition = 33; //centred.  For programming purposes rudder is assumed to move through 66 degrees lock to lock
```
sarfata: You have defined a constant `MAXRUDDERSWING` so you should probably do `MAXRUDDERSWING/2` here, no?

lib/KBox/src/tasks/IMUTask.cpp
sarfata: Looks awesome. Thanks a lot!

sarfata:
Rob,

Looks good to me. Sorry for the additional changes. If you don't have time, we can also go ahead and merge things as they are now and I will them.

I have answered on how to declare the `headingPID`, let me know if it does not work and I will do it for you and make a pull-request against your project ;)

Also, I do like using `RadToDeg` and `DegToRad`, I think it's worth it in the long run. 

I think we could simplify the way that the AutopilotPage communicates with the AutopilotTask a bit by having the page keep a reference to the task but I am not 100% sure of which variables I would pass by message and which I would just access directly so let's stick to what you have right now.

lib/KBox/src/tasks/ADCTask.cpp
matrixlinks: OK, removed separate Rudder Sensor task and created "useRudderSensor" boolean in ADCTask

lib/KBox/src/KMessage.h
matrixlinks: I tried changing the NAVMessage to store values as radians and adding the DegToRad and RadToDeg conversions on messages, but the PID code will no longer compile:
PID headingPID(RadToDeg(&apCurrentHeading), &apTargetRudderPosition, (RadToDeg)&apTargetHeading, (double)P_Param, (double)I_Param, (double)D_Param, (int)REVERSE);  error: cannot convert 'double*' to 'double' for argument '1' to 'double RadToDeg(double)'
I'm probably not doing this right.  For now I will leave it the way it was, which works.

lib/KBox/src/pages/NavigationPage.cpp
matrixlinks: I think making changes to the IMU code at this stage would interfere with how it sends N2K data.  I could be wrong about this but my impression from testing this with other N2K apps is that N2K sends heading data as positive values between 0 and 180 degrees and negative values (relative to 0) between 180 and 360.  So 30 degree compass is +30 (actually the radians equivalent), but 330 degrees compass is -30 (in radians equivalent), and the N2K software on the other end does the conversion to 360-30 = 330.  Am I off base here?

lib/KBox/src/KMessageNMEAVisitor.cpp
matrixlinks: Moved it inside the KMessageNMEAVisitor::visit function

lib/KBox/src/tasks/AutoPilotTask.cpp
matrixlinks: Tried that: won't compile with "error: no matching function for call to 'PID::PID()".  Evidently I'm not doing this right.  If you could take a crack at it that would be helpful.

lib/KBox/src/tasks/AutoPilotTask.cpp
matrixlinks: Added

lib/KBox/src/KMessage.h
matrixlinks: Not really certain what you are getting at here but have tried to make spacing consistent for everything on this page.

lib/KBox/src/KMessage.h
matrixlinks: Removed (missed these ones the last time through)

lib/KBox/src/tasks/AutoPilotTask.cpp
matrixlinks: In particular, I'm also not sure how I would use this code.  Does this constructor create the HeadingPid object that can be accessed in other functions on AutoPilotTask.cpp?

One of the issues I've encountered here and elsewhere is that I can only send KMessage objects from the "loop" function in any task.  If I try to do "sendMessage" from any other function nothing happens.  So I'm stuck on how to create a headingPID object, and access that object in both setup() (assuming we would like a distinct setup function) and the loop.

I'm having the same issue with the MFDTask code:  when the MFDTask::monitorN2kMessage gets an N2K message, I can extract the data, but if I try to create a MONMessage (the KMessage I'm using to send position, label and value to the display), I can't send it from within the monitorN2kMessage function.  I can only send it from the loop function in MFDTask.  Nor can I figure out how to create a MONMessage object inside the monitorN2kMessage function and then pass the object to the loop function to send the data to the display page.  Obviously there are some very different ways that c++ deals with objects than what I am used to! Any insights or practical solutions you can offer would be great.   The easiest solution here would be if I could send the Kmessage from within the function.  There's probably a switch somewhere in the visitor code structure that could let me do that.

matrixlinks:
Hi Thomas:  Subject to any further minor edits, I think I've done what I can with this, and have uploaded the latest edits.

 As a java guy I'm gradually starting to get my head about some of the big differences with c++, but I'm still stymied on how to actually make the PID object declaration work.   Looking forward to your insights as to how to best move forward.

This is interrelated with the RadToDeg and DegToRad conversions between the AutoPilotTask and the NavigationPage.  Not sure how to implement the conversion in relation to the "&variable" references in the headingPID.

In addition, I'm not sure what happens next with respect to creating a pull request for the MFD functionality once we're finished (for now) with the autopilot.  Do I create another branch?  Or can I use the same branch to create another pull request once the autpilot request has been processed and I've uploaded the new MFD code?

Thanks for your patience as I try to figure all this out!

Rob

lib/KBox/src/pages/NavigationPage.cpp
sarfata: Sorry - Yes you are right, I went with the NMEA2000 convention here and we should stick to it.

lib/KBox/src/KMessage.h
sarfata: What I meant is that there are a few white spaces at the beginning of an empty line (github highlights them). Not a big deal, I will take care of that.

lib/KBox/src/tasks/AutoPilotTask.cpp
sarfata: I have done the following:
 - moved setup code to setup, removed `isSetup`
 - moved `headingPID` inside the `AutoPilotTask` so now it is available in `loop()` and in `setup()`

Regarding your question about the MFD: the MFD class should have variables that it updates when it receives nmea messages. And then whenever it runs the loop, it can update the display. You can look at the `BatteryMonitorPage` for an example, it's very similar. Sorry if I am not understanding the question.

sarfata:
Hey Rob, so I have made some changes that I have pushed. If you look at this pull request in github, you can see each of my commits and see what I have changed. Feel free to ask me questions about anything.

There is still a few things that I think we can improve to make the code easier to read and maintain in the future. I will add more comments, please tell me what you think.

lib/KBox/src/KMessage.h
```diff
@@ -191,6 +196,94 @@ class IMUMessage: public KMessage {
     };
 };
 
+class NAVMessage: public KMessage {
+  private:
+    bool apMode, apHeadingMode, apWaypointMode, apDodgeMode;
+    double currentHeading, targetHeading, courseToWaypoint;
```
sarfata: I don't think the `AutopilotTask` should rely on `currentHeading` being provided by the `NAVMessage`. If the currently displayed page on KBox is not the navigation page, then those messages will not be sent and the autopilot will not get updates to the heading.

I think that instead, the autopilot should listen to `IMUMessage` and update its currentHeading based on IMUMessages.

Do you agree? If yes, I can do it - or you can do it. Tell me what you prefer.

Also, I don't think there is a difference between targetHeading and courseToWaypoint. I think they are both the same. As far as I can tell, `courseToWaypoint` is not used now and we should remove it so it's easier to read and understand the code.

sarfata: Finally, I think your code would greatly benefit if you replaced your 4 mode variables with one enum that lists all the possible states of the autopilot. There are so many variables now that it is hard to verify that we are not missing a case in the `if` statements that do logic on them.

lib/KBox/src/pages/NavigationPage.cpp
```diff
@@ -0,0 +1,280 @@
+/*
+  The MIT License
+
+  Copyright (c) 2016 Thomas Sarlandie thomas@sarlandie.net
```
sarfata: You should put your name here ;)

lib/KBox/src/pages/NavigationPage.cpp
```diff
@@ -0,0 +1,280 @@
...
+        }
+    } else {
+        apModeString = "Calibrating   ";
+        apModeDisplay->setText(apModeString);
+        apModeDisplay->setColor(ColorRed);
+    }
+    
+    rawHeading = RadToDeg(imu.getCourse());
+    if (rawHeading < 0){
+        navCurrentHeading = 720 + rawHeading;//adding 720 for nav math purposes
```
sarfata: This part of the code is pretty hard to understand and will be hard to debug.

Some ideas to make things easier here:
 - Write a little utility function `formatAngle()` that takes an angle in radian and returns a value in degrees between 0 and 360. Then reuse this everywhere you want to display an angle. You could have two if you want one that returns something between 0 and 360 and another one that returns values between -180 and +180.
 - Don't do navigation math in the navigation page. If the navigation page is not displayed then this code will not get executed. The autopilot needs to do all the navigation math. (If you follow my advice above to remove `navCurrentHeading` from `NAVMessage` then you will not need that math here and this function will just have to worry about displaying things.
 - Use only radians for the navigation math. You should only convert radians to degrees when displaying to the user.

lib/KBox/src/pages/NavigationPage.cpp
```diff
@@ -0,0 +1,280 @@
...
+        navCurrentDisplayHeading = 360 + rawHeading;
+    } else {
+        navCurrentHeading = 720 + rawHeading;
+        navCurrentDisplayHeading =  rawHeading;
+    }
+    headingDisplay->setText(formatMeasurement(navCurrentDisplayHeading, ""));
+    headingDisplay->setColor(colorForCourse(navCurrentDisplayHeading));
+    
+    if (apMode == true && apHeadingMode == true ){
```
sarfata: If you make `apMode` an enum then you can replace all those `if()` by one `switch (apMode)` statement. It will be much easier to read.

lib/KBox/src/pages/NavigationPage.cpp
```diff
@@ -0,0 +1,280 @@
...
+                navTargetDisplayHeading = 360 + (navTargetHeading - 720) ;
+            } else {
+                navTargetDisplayHeading = navTargetHeading - 720;
+            }
+        }
+        targetHeadingDisplay->setText(formatMeasurement(navTargetDisplayHeading, ""));
+        targetHeadingDisplay->setColor(colorForCourse(navTargetDisplayHeading));
+        
+        //following code recalculates navCurrentHeading based on "offcourse" value to avoid non-linear heading numbers across 180 degrees
+        navOffCourse = navCurrentDisplayHeading - navTargetDisplayHeading;
```
sarfata: It looks like `navOffCourse` is used only in this function so it does not need to be declared in the class. You can just declare it in this function.

sarfata:
Oh and for the MFDPage you can work on another branch. Make sure you start it from `master` (so you will not have the autopilot for now) and we can start discussing it in parallel.

lib/KBox/src/KMessage.h
matrixlinks: I can see your point on having the AutoPilotTask getting the current heading directly from the IMU message.  I'll tackle that.

We can take out courseToWaypoint if you like.  It was there as a placeholder for the next stage in the AP which would be adding the ability to navigate to a waypoint provided from a plotter.  Which raises a whole other conversation ... probably should have it elsewhere, so I'll put it on the discussion list ... which is can we get a message back over WiFI from a mobile device running nav software ?

With respect to replacing the mode variables with an ENUM, I thought about that when you suggested it before and started to do it, but then I realized there were issues with that.  Let's say we have 4 ENUMs that are mutually exclusive: OFF, HEADINGMODE, DODGEMODE, WAYPOINTMODE. Let's say we're in WAYPOINTMODE and we go into DODGEMODE to avoid a windsurfer.  When we come out of DODGEMODE, how does KBOX know what mode to go into then?  Does it go back to HEADINGMODE or WAYPOINTMODE?  We wouldn't know unless we track another various like previousMode. Also, it means that when we go into DODGEMODE, KBOX can keep the nav calculations going for either HEADINGMODE or WAYPOINTMODE in the background and resume giving appropriate steering directions when dropping out of DODGEMODE. Given these scenarios, having thought about it, I decided it would be cleaner to have independent boolean variables than going with your ENUM suggestion.  I'm happy with that decision, but would be interested if you think there is a cleaner solution now that you understand more about why I made the choice I did.

lib/KBox/src/pages/NavigationPage.cpp
matrixlinks: OK

lib/KBox/src/pages/NavigationPage.cpp
matrixlinks: Have already explained my logic for the independent booleans

lib/KBox/src/pages/NavigationPage.cpp
matrixlinks: I'll take a look at that approach.  As you know this is all new to me, but I'm feeling like it would have been a bit more constructive to get design suggestions like this a bit earlier in the review process:  I'm now starting to feel like I'm going backward.

lib/KBox/src/pages/NavigationPage.cpp
matrixlinks: OK

lib/KBox/src/pages/NavigationPage.cpp
matrixlinks: So have thought a bit more about this.  I can see the logic for moving the currentHeading  / offCourse calculations into the autoPilotTask.  We still need to determine targetHeading in the navPage because it depends on encoder events that are only detected in the navPage.  So we still need to do the targetHeading math in the Nav page.  

I'm not sold on doing all the math in radians.  Everybody using this code is going to need to do some customization for their particular boat in determining what adjustments to use for MAXRUDDERSWING, deadzone, slack and the PID variables.  I don't think anybody will want to try and figure out what their MAXRUDDERSWING is in radians, or try and tune the AP performance based on radians.  I think that would be counterintuitive for just about everybody.  So I don't see any practical advantage to doing all the calculations in radians.  Am I missing something?

I'll go ahead and move the currentHeading calcs into the AutoPilot, but will hold off on doing everything in radians, pending a persuasive argument why that has practical, rather than purely theoretical benefits.

matrixlinks:
Thanks for the clarification on how to move the MFD stuff forward,  I have a couple of technical challenges I don't yet know now to solve.  Once I've made the next round of AP updates,  I'll create a separate branch, clean up the MFD code, and then do a pull request and inside the code I will have some comments to clarify the issues I'm struggling with.

Thanks!

lib/KBox/src/KMessage.h
sarfata: > We can take out courseToWaypoint if you like. It was there as a placeholder for the next stage in the AP which would be adding the ability to navigate to a waypoint provided from a plotter. 
I understand but I think it's best to keep the code as simple as possible. It will be easy enough to add it when we add support for waypoint. Right now, I think it just makes things more complicated to read and understand.

> Which raises a whole other conversation ... probably should have it elsewhere, so I'll put it on the discussion list ... which is can we get a message back over WiFI from a mobile device running nav software ?

Not yet but I have been working on a big branch that has a ton of changes and that is one of the very important ones. Wifi and host will be able to exchange different type of messages (status of the connection, incoming frames in signalk or nmea format, etc).

> With respect to replacing the mode variables with an ENUM, I thought about that when you suggested it before and started to do it, but then I realized there were issues with that. Let's say we have 4 ENUMs that are mutually exclusive: OFF, HEADINGMODE, DODGEMODE, WAYPOINTMODE. 

What you are building here is a state machine. This is a perfect fit for what you are doing here but the risk when writing them is bugs in the transition from one state to the other (for example, forgetting to update one of the field on the display).
I will take a stab at rewriting this in a slightly different way and hopefully I can show you that it makes things easier to read and understand.

sarfata: Can you help me understand what is DODGE MODE? 

Is it a mode where we stop controlling the motors for X seconds so that the helmsman can dodge something else and then quickly get back to the previous setting without having to re-set the right heading?

matrixlinks:
Hi Thomas:

Yes, its common on Autopilots.  A seadoo is on a collision course and though you are tempted to hit it, you decide not to, go into dodge mode, manually steer for a few seconds to avoid, then go back out of dodge mode and resume the previous heading.  Or a series of waves pushes you off course and you decide to give the AP a break, go into dodge mode, steer back onto the proper heading, then go back out of dodge mode to resume normal AP operation.  Before going into Dodge mode we could have been in Heading Mode or Waypoint mode.  In Dodge mode, we stop the AP controlling the rudder, although the device continues to calculate where the rudder should be.

lib/KBox/src/KMessage.h
matrixlinks: Yes, its common on Autopilots.  A seadoo is on a collision course and though you are tempted to hit it, you decide not to, go into dodge mode, manually steer for a few seconds to avoid, then go back out of dodge mode and resume the previous heading.  Or a series of waves pushes you off course and you decide to give the AP a break, go into dodge mode, steer back onto the proper heading, then go back out of dodge mode to resume normal AP operation.  Before going into Dodge mode we could have been in Heading Mode or Waypoint mode.  In Dodge mode, we stop the AP controlling the rudder, although the device continues to calculate where the rudder should be.

lib/KBox/src/tasks/AutoPilotTask.cpp
```diff
@@ -0,0 +1,132 @@
...
+  headingPID.SetOutputLimits(0.0, MAXRUDDERSWING); //output limits  0 = full starboard rudder (trailing edge of rudder to the right, bow moves to right)
+  headingPID.SetSampleTime(AUTOPILOT_SAMPLE_TIME);
+  DEBUG("Init autopilot complete");
+}
+
+void AutoPilotTask::loop() {
+  headingPID.Compute();
+
+  apRudderCommandSent = apTargetRudderPosition; //by default send rudder command to equal target position, then modify it
+  double testDegs = apTargetRudderPosition - apRudderCommandSent;
```
sarfata: I don't understand these two lines. The way they are right now, it seems that `apRudderCommandSent` will always be equal to `apTargetRudderPosition`, and so `apTargetRudderPosition - apRudderCommandSent` will always be 0 and the following `if` block is never used.

I am guessing you meant something else here.

lib/KBox/src/tasks/AutoPilotTask.cpp
```diff
@@ -0,0 +1,132 @@
...
+  sameLastDirection=true;
+  apTargetRudderPosition = MAXRUDDERSWING / 2; //centred.  For programming purposes rudder is assumed to move through 66 degrees lock to lock
+  headingPID.SetMode(AUTOMATIC);
+  headingPID.SetOutputLimits(0.0, MAXRUDDERSWING); //output limits  0 = full starboard rudder (trailing edge of rudder to the right, bow moves to right)
+  headingPID.SetSampleTime(AUTOPILOT_SAMPLE_TIME);
+  DEBUG("Init autopilot complete");
+}
+
+void AutoPilotTask::loop() {
+  headingPID.Compute();
```
sarfata: Do you think the PID is smart enough to understand that to go from 358 to 005, it needs to go right and not left? I think this will require some special attention in the PID algorithm, or maybe some pre-processing of the inputs.

lib/KBox/src/KMessage.h
sarfata: I did some research on how autopilot are defined by other protocols.

- In [NMEA0183](http://catb.org/gpsd/NMEA.html#_apb_autopilot_sentence_b) the key sentence is APB which is sent by a GPS or a control unit and tells the pilot which heading to follow 
- Or [RMB](http://catb.org/gpsd/NMEA.html#_apb_autopilot_sentence_b) which is very similar. [Apparently that is the one OpenCPN uses](.
- [SignalK defines 8 possible "states"](https://github.com/SignalK/specification/blob/master/schemas/groups/steering.json#L33): "auto", "standby", "alarm", "noDrift", "wind", "depthContour", "route", "directControl"
- [SignalK defines 3 possible "target"](https://github.com/SignalK/specification/blob/master/schemas/groups/steering.json#L73) which are headingMagnetic, headingTrue, windAngleApparent
 - [This SignalK plugin](https://github.com/sbender9/signalk-raymarine-autopilot/blob/master/index.js) has NMEA2000 commands to control a Raymarine autopilot. The four supported states are auto/route/wind/standby.

There are a few things I draw from this research:
 - Most current autopilots do not do navigation calculation to a waypoint. They rely on an external device (GPS unit, PC, iPad, etc) to tell them what is the heading to destination and they follow it.
 - It would be really cool to display this information on KBox so that whether you have a human pilot or an autopilot, you could see on KBox how well you are doing towards your waypoint, whether you are on course and how far you are.
 - The way to "dodge" with a traditional autopilot is to press -10/+10 (see [this discussion](http://www.cruisersforum.com/forums/f134/avoid-function-with-opencpn-and-autopilot-140503.html)). This could be done by just turning the rotary button on KBox.
 - We could have an extra "follow waypoint" feature where KBox would be smart enough to be both the autopilot control head and the navigation system. We would have to include also detecting when you get in the arrival circle or the perpendicular of the waypoint which I think they do by keeping the original heading to target. If we have this feature, dodge makes a lot of sense.

So coming back to our current pull request :) All this research helped me realize that what we are doing here is "just" the autopilot head control. Not the navigation part. I agree we want to do the navigation part at some point, but to do that we will need to be able to parse incoming messages to get current position and position of a target waypoint. Let's build things one step at a time.

So this is what I suggest and will push in a different branch for you to review:
 - Rename NavigationPage to AutopilotControlPage, it will be dedicated to controlling an Autopilot from KBox.
 - Remove reference to waypoint for now, focus on having a targetMagneticHeading and building an autopilot that can follow it.

matrixlinks: I can guarantee from my experience with autopilots that have a dodgemode that just changes the targetheading but maintains control of the steering would be a bad idea.  Some autopilots have a "powersteering" mode that behaves similar to how you describe press -10/+10.  I would not use an autopilot that did not give me the option to temporarily deactivate the steering commands.  

And yes, just maintaining a set heading was my initial intention for version 1.  I thought it might he helpful to others and a hint to ourselves to leave some breadcrumbs in the code for future intended features like getting a heading to a waypoint, but that does not seem to be your preference.  The name of the page is immaterial.

lib/KBox/src/tasks/AutoPilotTask.cpp
matrixlinks: I think line 77 was originally intended to be in the setup function.  I moved it there and everything still functions properly.

lib/KBox/src/tasks/AutoPilotTask.cpp
matrixlinks: You should know from testing the AP code on KBOX that it works!  That's the magic of the math approach that Robert Huitema developed.  Circular compass math is converted to to linear math by adding 720 and defining the heading data that feeds the PID function as targetheading plus offcourse.  If you set DEBUG on the numbers being fed into the PID you'll see how this works.  In effect, the inputs are already "preprocessed" as you suggest.

Have taken a brief closer look at moving the math functions from the current Nav page into the AP task.  That's going to take more time than I have available for the next few days.  I'll look again later in the week.  I'm not entirely sure at this point whether the practical benefits of doing so are worth the effort, particularly as I don't think we can entirely eliminate the math in the Nav page in any event.

Issue #10, opened 2016-09-09
Add a page showing realtime IMU data


sarfata:
So that users can look at it in realtime and make sure everything is working well.

The IMU should show:
- Current pitch
- Current yaw
- Current roll
- Current magnetic course

CaptainRon47:
My suggestion is to have the KBox display pages cycle thru several pages on a timed cycle (releasing the push button for other options).  It could show the current two pages plus pages for all instrument data (wind, barometric pressure, IMU data, etc).

sarfata:
(removing barometer from this ticket - i will create another one for that)

PR #30, opened 2017-05-24
Misc housekeeping

neftaly:
This PR has several tiny changes:
* Replace N2K numerical PGNs with their text aliases (for readability)
* Send heading data over Wifi (for openCPN etc)
* Use `undefined` instead of `42` for humidity value

### DIFF
--- a/lib/KBox/src/converters/BarometerN2kConverter.cpp
+++ b/lib/KBox/src/converters/BarometerN2kConverter.cpp
@@ -33,7 +33,10 @@ void BarometerN2kConverter::visit(const BarometerMeasurement &bm) {
   static int sid = 0;
   // Seems the i70 display will only show sea or outside temperature and
   // outside humidity.
-  SetN2kPGN130311(n2km, sid++, N2kts_OutsideTemperature, CToKelvin(bm.getTemperature()),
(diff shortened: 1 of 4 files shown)


neftaly:
#10

sarfata:
Thanks @neftaly ! Looking great!

I will merge this now. The 42% humidity should have been removed a long time ago. We might also remove the temperature because the temperature inside the box is always significantly higher than the temperature outside the box.

I have been doing very significant changes in the `develop` branch and associated PRs (#31 for example). I have not merged yet in master to avoid confusing people but if you are going to do more work on KBox, I would strongly suggest you work from `develop` instead of `master`.

Thanks!

neftaly:
Ok, sweet!

neftaly:
How about using `N2kts_InsideTemperature`? Later hardware could use something like an external BME280.

sarfata:
@neftaly sorry I had missed your message

> How about using N2kts_InsideTemperature? Later hardware could use something like an external BME280.

Because the i70 display only displayed the outside temperature. Very selfish reason ;) This should be inside, I agree and this should be configurable. It's so far off from the actual inside temperature that I think I have stopped sending it anyway.

PR #37, opened 2017-07-03
Features/sktool

sarfata:
Adds a tool 'sktool' that can be used to convert NMEA0183 sentences to signalk messages.

This makes it very easy to test KBox code with different nmea sentences and see how they are converted.


src/common/signalk/SKJSONVisitor.cpp
```diff
@@ -30,16 +30,47 @@
 
 #include "SKJSONVisitor.h"
 
+void SKJSONVisitor::processSource(const SKSource &source, JsonObject &sourceObject) {
+  sourceObject["label"] = source.getLabel().c_str();
+
+  switch (source.getInput()) {
```
neftaly: Any chance the input pin ID could be included in the signal K output JSON (maybe in the metadata)?

sarfata: @neftaly Thanks for following the PRs!

The way it's done right now you will get the exact input on which data was received (NMEA2000, NMEA1, or NMEA2). 

Internal sensors like IMU, Barometer and ADCs will appear as coming from KBox. 

What kind of extra precision are you looking for? Can you give me an example?

neftaly: I got the impression that "talker" was supposed to identify the source of Signal K data itself (i.e. always the kbox), and not who originally generated the NMEA data, so that you could tell multiple sK streams apart? I probably misunderstood that, in which case this is fine :)

PR #47, opened 2017-09-30
Fix IMU units discrepancies

### LINKED ISSUE #46: N2k Message for IMU in degrees instead of radians
in the NMEA2000Task.cpp seems to be a bug, as there is sent degrees to form N2k-Sentences.

Shouldnt be there a DegToRad as:
```
SetN2kAttitude(n2kmessage, _imuSequence, DegToRad( m.getYaw() ), DegToRad( m.getPitch() ), DegToRad( m.getRoll() ) );
    sendN2kMessage(n2kmessage);

    double magneticVariation = N2kDoubleNA;
    double magneticDeviation = N2kDoubleNA;

    SetN2kMagneticHeading(n2kmessage, _imuSequence, DegToRad( m.getCourse() ), magneticDeviation, magneticVariation);
```
Tested with Actisense NGT-1 and NMEA Reader shows correct values then.

sarfata:
- IMU sends data in degrees (code assumed radians)
 - NMEA2000 expects radians
 - NMEA expects degrees
 - We should always store in radians internally

This fixes #46.

### DIFF
--- a/lib/KBox/src/KMessageNMEAVisitor.cpp
+++ b/lib/KBox/src/KMessageNMEAVisitor.cpp
@@ -63,20 +63,23 @@ void KMessageNMEAVisitor::visit(const NMEA2000Message &n2km) {
   free(s);
 }
 
+
+inline static double RadToDeg(double v) { return v*180.0/3.1415926535897932384626433832795; }
+
 void KMessageNMEAVisitor::visit(const IMUMessage &imu) {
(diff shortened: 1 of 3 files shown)


sarfata:
@ronzeiller do you want to take a look before I merge this?

I have done a similar fix on the develop branch. That was a very serious bug! Thanks for pointing this out!

PR #54, opened 2017-11-06
KBox with Teensy36

ronzeiller:
Sarfatas develop branch forked.
New branch teensy36 
Updated some libraries: ADC, FlexCAN (Timo) and SDFat to work with Teensy 3.6 and built in SD-Card
Platformio.ini splitted for compiling and uploading just the code for teensy without ESP

Should work with KBox, but may be better to merge it into an own branch?
There should follow some code with additional options for KBox....


sarfata:
Thanks @ronzeiller - This all looks awesome and I would like to get into the habit of regularly merging your changes with my work so we can effectively collaborate.

I think the best way to do this would be to have pull-requests for specific features so we can discuss them instead of putting everything in one big PR where it will be very hard for me to follow and test on KBox.

I have a few questions/comments:
 - How did you update the libraries? Did you use the git subtree commands listed in lib/README.md? I checked out your branch and tried to update ADC again, this gave me lots of conflicts which is a sign that maybe you just dropped the file in and commited them. We should avoid this, it will make harder to stay in sync with the parent project.
 - I think we can have one platformio.ini file that would work for your project and would continue to work with the standard KBox. You could have just added a new target "teensy36". Why did you remove the rest?
 - I am very interested in your code to configure the KBox. Why did you choose a .ini file? We already have a JSON parser in KBox, we could re-use it. In any case, I would like to get this into the main KBox.
 - I am also very interested by the IMU page and the axis remap function. Thanks for the effort on this.

I think there are two ways we can go forward:
 1 - You can break this branch into different pull-request and we can look at them one by one, discuss them and merge them
 2- I take code from your branch, create new pull-requests, you review them and then we merge them.

If you are not super familiar with Git, the second option might be easier for you. I would not mind doing this.

In both case, the first step would be to make a new PR to update the libraries and only do this. It will make it easier to look at the rest of the files that have changed.

ronzeiller:
_I think the best way to do this would be to have pull-requests for specific features so we can discuss them instead of putting everything in one big PR where it will be very hard for me to follow and test on KBox._

Thank you, I am glad that you are interested!
Please take into account, that I am quite a novice in github.
So the first pull request was just small add ons for Teensy 3.6.
I was not aware of that the whole code I changed later on was going to the same PR.


_I have a few questions/comments:_

_How did you update the libraries? Did you use the git subtree commands listed in lib/README.md? I checked out your branch and tried to update ADC again, this gave me lots of conflicts which is a sign that maybe you just dropped the file in and commited them. We should avoid this, it will make harder to stay in sync with the parent project._

Also sorry for that. I read it on the Readme, but was not able to install it in my atom/platformio.
And I saw, that some of the libraries are changed by you in your project, therefor I was very carefully with automatic github library updates.

On the ADC library: I took it from Github on 14th of September and as you say, I put it in the library folder of my project, because the version of KBox firmware was not working with the Teensy 3.6



_I think we can have one platformio.ini file that would work for your project and would continue to work with the standard KBox. You could have just added a new target "teensy36". Why did you remove the rest?_

OK, you mean it is possible just to add another board?
With the rest in the platformio.ini everything else was always uploaded too. 
So it was a workaround for me to just upload the `[env:host]` part.
May be I should upload just the `[env:host]` by CLI?  But I do not know how :-)

_I am very interested in your code to configure the KBox. Why did you choose a .ini file? We already have a JSON parser in KBox, we could re-use it. In any case, I would like to get this into the main KBox._

When I began with the ini-File I was working on the KBox master branch (without the JSON).
Now I am in the develop branch and realise that a lot of nice things are implemented already.
The ini-File is a very, very simple to edit thing. Much simpler than any JSON, XML or similar.
But, the end scenario will definitely be a web browser based configuration and then JSON will be a real advantage.
So I think ini-File is probably just an easy and quick solution for now.
Also for getting an oversight: how many things are necessary for configuring?


`I am also very interested by the IMU page and the axis remap function. Thanks for the effort on this.`

The disadvantage of the IMU-Page as it is at the moment:
The interval needed for Autopilot is too much for a display. So I put the interval down to 100, but this is no solution.
BTW, I was also experimenting with the encoder-button. A long button-click was storing the calibration values into the EEPROM. The aim was also to store an offset for Zero-Setting of the well balanced boat before you go out sailing.....

_I think there are two ways we can go forward:
1 - You can break this branch into different pull-request and we can look at them one by one, discuss them and merge them
2- I take code from your branch, create new pull-requests, you review them and then we merge them.
If you are not super familiar with Git, the second option might be easier for you. I would not mind doing this.
In both case, the first step would be to make a new PR to update the libraries and only do this. It will make it easier to look at the rest of the files that have changed._

No, I am not super familiar with git at all :-) 
Whatever you think would be best, I will follow

sarfata:
> The disadvantage of the IMU-Page as it is at the moment: The interval needed for Autopilot is too much for a display. So I put the interval down to 100, but this is no solution.

What do you mean? You want to send data faster or slower to the autopilot?

> re: Platformio configuration

You should copy the 'host' section to a 'teensy36' section and then when you compile in platformio, in the bottom left corner, only select 'teensy36' to only compile this one.

I have screenshots of this here: https://github.com/sarfata/kbox-firmware/wiki/Developer-Setup#using-atom-to-build-and-flash.

ronzeiller:
I meant, that it was just for the moment to put down the frequency for testing the display.
The high frequency will be important for the Autopilot I guess.
Internal calculations should/could be on the higher frequencies of getting datas from the bus (or sensors).  If the datas are used for performance calculations by the tactic_pi in openCPN, then more datas would probably better too. But outputs for let's say "normal" navigation in openCPN or output just for displays could be lower, e.g. around 1Hz. So it depends much on the use of the output.
Short answer: I think that the KBox display of sensor- and boat-datas would need damping.

Thanks for the tip with selecting the PIO build!

ronzeiller:
> How did you update the libraries? Did you use the git subtree commands listed in lib/README.md? I checked out your branch and tried to update ADC again, this gave me lots of conflicts which is a sign that maybe you just dropped the file in and commited them. We should avoid this, it will make harder to stay in sync with the parent project.

At the moment I am testing to install some of the libraries with platformio's library manager.
The big advantage is, that you get a message when the library is updated to a new version, which you do then with one click.

With the subtree and git it seems all a little bit more complicate?

The disadvantage of this method: the libraries are hidden from the project and not so easy to search for something.
Also the libraries which are customized with DEBUG and similar cannot be used with automatic updates.

BTW: new ADC library gave me a lot of errors too.
But the Platformio framework-arduinoteensy is a little bit out of date by it's version 1.36 (Teensyduino just got an update to 1.4) 
Copying the core files from Teensyduino into the framework-arduinoteensy helped.

sarfata:
I have used libraries more in the beginning. The big problem with libraries is that I can never be completely sure that someone else will get exactly the same version of the library than me. This is a big big problem because it might compile for you but not for me.

So in the end, I have found that having a copy of the file that I am interested in in the project allows me to have complete control on what version of the library people will use. Yes it does mean that developers have to learn how to use git subtree but it's really not that hard and worth it in the end I think.

In any case, it was a great idea to make NMEA2000 a library! Now will you make a PR for the NMEA2000_teensy library ;) ?

ronzeiller:
just sent a PR for making NMEA2000_Teensy a library too.

For other users it is perfect, when they find the library which was compiled correct and they can download the whole thing.
I was very happy with this too!!!


Not so hard to make the subtrees (you describe it very good) but hard to alwasy keep them updated during development.

The problem might be that people have already installed newer libraries for their own projects.
or have newer versions of Platformio or Teensyduino, just like we had with the ADC library?

May be, when all the relevant libraries are installable by Platformio it will make things easier :-)

ronzeiller:
What do you think about, you make a new branch (e.g. teensy36)  from your develop branch. 
Then I could do a rebase or a new clean fork and I could PR small stepped changes to this new branch, which we can discuss?

ronzeiller:
There are too many updates meanwhile in my code. 
There are also updates for NMEA2000 Library, ADC Library, ILI9341, Teensyduino resp. a brand new one for Platformio (Version 2.7.0)

What do you think about, you make a new branch (e.g. teensy36) from your develop branch, after updating libraries and IDE.
Then I do new, clean fork on that and I start from the scratch with small changes to this new branch, which we can discuss?

If you are interested in my work on the Teensy KBox, then you are welcome to look at 
https://github.com/ronzeiller/kbox-firmware-teensy36.git

PR #55, opened 2017-11-06
Bugfix in Develop branch

ronzeiller:
I think there is/was a bug in the rxValidEvent and rxErrorEvent Counter in NMEAService.cpp

.event --> counter++
.countEvent --> get the count value


src/host/services/NMEAService.cpp
```diff
@@ -149,8 +150,9 @@ void NMEAService::loop() {
   DEBUG("Found %i sentences waiting", receiveQueue.size());
   for (LinkedList<NMEASentence>::iterator it = receiveQueue.begin(); it != receiveQueue.end(); it++) {
     if (it->isValid()) {
-      KBoxMetrics.countEvent(rxValidEvent);
-      this->sendMessage(*it);
+      KBoxMetrics.event(rxValidEvent);
+      // DEBUG( "Rx NMEAValidCounter: %i", KBoxMetrics.countEvent(rxValidEvent) );
```
sarfata: Can you remove this line? WE are not going to need it.

lib/readme.txt
```diff
@@ -0,0 +1,36 @@
+
```
sarfata: I don't think we should add this file. There is already a readme in the lib folder which has more important and more relevant content.

src/host/services/NMEAService.cpp
ronzeiller: debug message is deleted.
the readme.txt was made by platformio automatically (always restored when I deleted it)
I will have to find a way to experiment and try, including debug messages for testing without "damaging" your code....

sarfata: It looks like you did not commit the change to remove that debug message. Can you remove it?

sarfata:
Thanks @ronzeiller - This is now merged. We should use the same system for every new feature but I will help you.

PR #60, opened 2017-11-22
Feature/update libraries

sarfata:
Switch from including libraries in the source tree to letting platformio pull them from github. This will make updating libraries a lot easier in the future.

@ronzeiller I think this will make you happy!


ronzeiller:
@sarfata yes, you made me happy :-)

Just made a fork and the only thing I had to do for Teensy 3.6 was to change the "board" in platformio.ini and everything was compiling perfect

Perfect!!!
Great work!!!

PR #61, opened 2017-11-23
Changes submitted by Ron aka rescotti47

sarfata:
This includes all the changes that Ron did for his own boat. I will be using this to make sure the develop branch does everything Ron needs:

- [ ] Enable the NMEA outputs
- [ ] Set NMEA2 speed to 4800 bauds
- [ ] Change the unit for pressure when exporting in NMEA format
- [ ] Fix the NMEA sentence for roll/pitch to work with OpenCPN
- [ ] Allow user to configure orientation of KBox
- [ ] Allow user to disable HDM output (magnetic heading)
- [ ] Display barometer pressure on screen
- [ ]


CaptainRon47:
I am not sure how I can help you with this, all of my changes were adhoc and not conforming with any coding rules (oh well, they worked for me). .  It looks like you are already addressing many of my changes in the IMU branch (at least the ones with IMU setup, new display page and location of 'constants').   This PR could address how tell Kbox what data to write to the various outputs. My thoughts would be that somewhere in the new config repository we set flags that would define what gets output at; the USB, the SD card, the WiFi and the NMEA outputs.  So for me the N2K_USB_Flag would be off.  I also turned off DeBug, so the DeBug_USB_Flag would be off. while the NM83_USB_FLag would be on.  At least that would be my naive way of doing it.  
I guess I will start by locating all my changes in the current (12/13/17) Master.

CaptainRon47:
Well, I have started to go thru the new Master and while I am pretty confused I am picking up some things.  I see that the encoder switch does have rotation options (A/B) not just the  push button, you have gone to a logging approach for DEBUG, INFO and ERROR (but I don't yet understand how this works).  I also found in the MFG section the NMEATest.h which shows me that to turn on the NMEA1/2 outputs i need to put a '1' in the digitalWrite statement.  I tried that in my edits, but never did see any output on those pins.  

I feel a bit like I am back at square one for making my changes, but at least I have my previous build which is doing more or less what I wanted it to do.  I am still not sure how I can help you with rewriting this code, but at least you have my inputs.

sarfata:
@CaptainRon47,

I created this pull-request so I have an easy way to look at all the changes you had done and make sure we cover all of them in the new master. I was not expecting you to go do the same changes again, I will do that for you. As you said, the big value here is showing me what is useful for you! And showing me how you solved it which always makes things a bit easier to do.

thomas

CaptainRon47:
Thanks, I was just trying to see if I could help, but I am afraid I am out of my league (at my current level of C++ understanding) for writing modern efficient code!  So I will be patient and wait and see where you go with it.

Issue #74, opened 2017-11-23
Save accelerometer/gyro/compass calibration and restore it when rebooting


sarfata:

ronzeiller:
I have made a routine for this, I will make a PR, have to look for it.

It did not help me to get proper values, but I have the box on my desktop and without moving the sensor the values for HDG are rather useless and wrong...

Heel and Pitch are reliable good and not drifting.
On one of my versions I had a long button press for storing offset values for heel/pitch to "calibrate" the boat in the marina, but it was for the master branch ;-)
To load those values from memory at rebooting is nice to have (especially when rebooting is necessary during sailing....)

sarfata:
Can you be more specific about "they are rather useless and wrong"? Did you connect the same sensor to your teensy? Is the value off when you compare it to a magnetic compass? Or completely bogus and jumping all over?

This sensor is supposed to be really good and is used in drones and many other applications. I think we can make it work very well. When I take the time to calibrate it at home (following the procedure in the datasheet), it gets very precise afterwards. I have not tried to calibrate and compare to a navigation compass on a boat yet.

sarfata:
I think we can close this ticket now. I agree we are not very happy with the result but the calibration is saved and restored.

PR #77, opened 2017-11-24
IMUMonitorPage with filtered values for HDG, Heel, Pitch

ronzeiller:
Filter is IIR-Filter as used in openCPN for damping of display in dashboard.

**What is missing and should/could be done:**
 - Setting for damping values by ini-file or similar 
 - for automatic damping (calculated by the IMUService interval) the real interval value should be taken (at the moment fixed value in main.cpp)
 - get cal-values from IMUService to show them too? 
- remove all the debug msg from the code after testing

**Note:** 
at the moment the display stucks to an old value, if there is no update due to bad calibration values!! (KBox minimum default SysCal 3 needed)
Workaround could be to check if there was no update since nnn ms and set the textlayers to something like "--"

**Remark:**
.....on the Display and in the code is mentioned "heel" and not "roll" as in KBox.
I would  suggest is to switch to "heel" as I think that "roll" is for large ship and "heel" is the more common expression for sailing boats. (formula for leeway, ORC calculations, openCPN Dashboard,....)

### DIFF
--- a/src/host/pages/IMUMonitorPage.cpp
+++ b/src/host/pages/IMUMonitorPage.cpp
@@ -1,0 +1,133 @@
+/*
+  * Project:  KBox
+  * Purpose:  IMU Monitor Page for HDG, Heel & Pitch (+ Cal planned)
+  * Author:   (c) Ronnie Zeiller, 2017
+  * KBox:     (c) 2017 Thomas Sarlandie thomas@sarlandie.net
+
+  The MIT License
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy
+  of this software and associated documentation files (the "Software"), to deal
+  in the Software without restriction, including without limitation the rights
+  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+  copies of the Software, and to permit persons to whom the Software is
+  furnished to do so, subject to the following conditions:
+
+  The above copyright notice and this permission notice shall be included in
+  all copies or substantial portions of the Software.
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+*/
+
+#include <stdio.h>
+#include "common/signalk/SKUpdate.h"
+#include "IMUMonitorPage.h"
+#include <KBoxLogging.h>
+#include "signalk/SKUnits.h"
+#include "services/IMUService.h"
+#include "common/signalk/SKUpdateStatic.h"
+#include "common/signalk/SKUnits.h"
+
+
+IMUMonitorPage::IMUMonitorPage(SKHub& hub) {
+  _IMU_HDGFiltered    = 361;
+  _IMU_PitchFiltered  = 181;
+  _IMU_HeelFiltered   = 181;
+
+  // Damping values 1.....100 --> 0.5 no filter ......0.005 max filter
+  // for higher frequencies even more
+  // TODO: make KBox setting for damping values 1.......100
+
+  // auto damping factored by the IMUServiceInterval which is
+  // probably between 50 and 100 (20 - 10 Hz)
+  // TODO: set and get IMUServiceInterval for automatic modus -> #define IMUServiceInterval ??
+
+  _dampingHdg = _IMUServiceInterval / 1400.000; // 0.05
+  if (_dampingHdg>0.5) _dampingHdg=0.5;
+
+
+  // Heel and Pitch need smaller damping
+  _dampingHeelPitch = _IMUServiceInterval / 700.000; // 0.1 .... 0.1;
+  if (_dampingHeelPitch>0.5) _dampingHeelPitch=0.5;
+
+  INFO( "HDG Damping factor taken for IIR-filter: %f", _dampingHdg );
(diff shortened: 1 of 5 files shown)


ronzeiller:
**Don't know why the Travis CI build fails with:**
**Seems that they have an old Teensy library!!
I had this error here before updating to Teensyduino 1.4 resp. before Update Teensy Module to 2.7 in Platformio**
```
In file included from .piolibdeps/Teensy_ADC/ADC_Module.cpp:36:0:
.piolibdeps/Teensy_ADC/VREF.h:23:38: error: 'VREF_SC_MODE_LV_HIGHPOWERBUF' was not declared in this scope
inline void start(uint8_t mo
de = VREF_SC_MODE_LV_HIGHPOWERBUF, uint8_t trim = 0x20) {
```

sarfata:
Yes I have seen that too. I am going to edit Travis.yml to force an update
before building.

Thanks for all the other PRs, i am going to look into it ASAP. Been busy
painting the boat these past few days ... different kind of project ;)

sarfata:
Unfortunately git thinks this was merged because your other PR included those commits at some point but they were reverted later so it does not appear in the current develop branch.

I think the best thing to do here is to take the files, create a new branch from develop, apply the files and open a new PR.

src/host/util/iirfilter.h
```diff
@@ -0,0 +1,87 @@
...
+* iirfilter.h
+*
+* Project:  KBox
+* Purpose:  Class that implements single order inifinite-impulse-response filter
+* Author:   Transmitterdan, adaptions for KBox by Ronnie Zeiller 2017
+***************************************************************************
+*   Copyright (C) 2016                                                    *
+*                                                                         *
+*   This program is free software; you can redistribute it and/or modify  *
+*   it under the terms of the GNU General Public License as published by  *
```
sarfata: This license is a problem. If we include a single GPL file then the entire project must be distributed under the GPL. This is very unfortunate but I think we have to be careful with this because it would be a shame to put so much effort into the project and get into legal issues later.

I think we should try to find a MIT or Apache licensed library that does the same thing. For example this would probably work well for us: https://github.com/JonHub/Filters

src/host/pages/IMUMonitorPage.h
```diff
@@ -0,0 +1,57 @@
...
+#include "common/signalk/SKSubscriber.h"
+#include "util/iirfilter.h"
+
+class IMUMonitorPage : public Page, public SKSubscriber {
+private:
+  TextLayer *hdgTL, *heelTL, *pitchTL, *calTL;  // display cal planned for later
+  unsigned long int _nextPrint = 0;
+  uint16_t _IMUServiceInterval = 50;
+
+  iirfilter   IMU_HdgFilter;
```
sarfata: I do not think you need to prefix the variable with `IMU_` since they are private member of this class. But they should start with a lower-case letter. Maybe `headingFilter`, `heelFilter` and `pitchFilter`?

src/host/pages/IMUMonitorPage.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  // Heel and Pitch need smaller damping
+  _dampingHeelPitch = _IMUServiceInterval / 700.000; // 0.1 .... 0.1;
+  if (_dampingHeelPitch>0.5) _dampingHeelPitch=0.5;
+
+  INFO( "HDG Damping factor taken for IIR-filter: %f", _dampingHdg );
+  INFO( "Heel/Pitch Damping factor taken for IIR-filter: %f", _dampingHeelPitch );
+
+  // set filter type
+  IMU_HdgFilter.setType(IIRFILTER_TYPE_DEG);
```
sarfata: The library I recommended does not support type degree unfortunately. We would need to add this.

src/host/pages/IMUMonitorPage.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  if ( up.hasNavigationHeadingMagnetic() ) {
+    const SKValue& vm = up.getNavigationHeadingMagnetic();
+    _IMU_HDGFiltered = IMU_HdgFilter.filter( round ( SKRadToDeg( vm.getNumberValue()) * 10) / 10.0 );
+    //DEBUG("HDG: %f\n", round( SKRadToDeg( vm.getNumberValue()) * 10) / 10.0 );
+  }
+
+  if ( up.hasNavigationAttitude() ) {
+    const SKValue& vm = up.getNavigationAttitude();
+    _IMU_HeelFiltered = IMU_HeelFilter.filter( round( SKRadToDeg( vm.getAttitudeValue().roll ) * 10) / 10.0 );
```
sarfata: I think it would be cleaner to do all the maths in radian and just convert when we are displaying the values.

sarfata:
> at the moment the display stucks to an old value, if there is no update due to bad calibration values!! (KBox minimum default SysCal 3 needed). Workaround could be to check if there was no update since nnn ms and set the textlayers to something like "--"

Yes that would be great!

> Remark:
> .....on the Display and in the code is mentioned "heel" and not "roll" as in KBox.
I would suggest is to switch to "heel" as I think that "roll" is for large ship and "heel" is the more common expression for sailing boats. (formula for leeway, ORC calculations, openCPN Dashboard,....)

SignalK went with `roll` for the spec, that is why I chose `roll` in the code. I have no problem with using `heel` for display but I do not think we should change the name in code or we will not be coherent anymore.

ronzeiller:
You are right about the license, I thought so too.
I will make a new file filters.h with redesigned iir-filter which we can extend with median-filters or other average filters as needed in the future.

Everything else you mentioned is totally fine. Will do everything new and make a new PR.
Thank you for your patience with all my github trial and errors.... :blush:

May be it would be good to split `update.setNavigationAttitude` and `update.setNavigationHeadingMagnetic` to "their own" calibration states  (Hdg -> magCalib and roll and pitch -> accelCalib and gyroCalib) in IMUService (and not to the general sysCalib)?

At the IMUMonitorPage I will look for Updates within the last, let's say 3 (?) seconds and make the display "--"
(Could/Should be done at any (monitor)-page, to get rid of old, not valid datas anyway......)

(More specific answer about the BNO055 sensor will come in the next days in Issues https://github.com/sarfata/kbox-firmware/issues/74)

sarfata:
> Maybe it would be good to split update.setNavigationAttitude and update.setNavigationHeadingMagnetic to "their own" calibration states (Hdg -> magCalib and roll and pitch -> accelCalib and gyroCalib) in IMUService (and not to the general sysCalib)?

I think I looked into that in the past but I believe I have seen the accel/gyro calib go to 2 even after sysCalib is 3 for a while. Was not great.

PR #80, opened 2017-11-25
More parsing of N2k PGN's

ronzeiller:
Started with some more NMEA2000 sentences (PGNs) to parse for discussion for continue to work on it:

1.) I am not sure if I can use for new PGNs: 
`SKUpdateStatic<2> *update = new SKUpdateStatic<2>();`
or need 
`SKUpdateStatic<3> ...`
or something else....

2.) changed if-statement  to switch as this should reduce CPU usage?

3.) added defined offset if not set in the Water Depth PGN
```
if (!N2kIsNA(offset)) {
      // When no offset then offset should be Zero
      offset = 0;
    }
```
4.) prepared (commented out) for config-settings if someone doesn't need a PGN converted 
As soon config or variables for that is more defined we could do something like this or  similar

```
//if ( cfEnableN2kToNMEA_127250 ) {
   return parse127250(input, msg, timestamp);
//}
```

5.) The parsing of Date & Time PGN 126992 could also be used for setting global true KBox System Date or Time.
I had this already working in on of my KBox branches ("old" master), which was also writing date/time for logging files on SD and automatically switched to a new logfile ad midnight.

I could start to implement this here by extending your SKTime class and add some (global) static variables. Or do you have a better idea may be?

Another way for getting true Date/Time is from NMEA GPS  (may be selecting this by config?)
(Should be selectable, as some users take KBox without any NMEA input and others wants it without NMEA2000 at all.....)

(A WiFi connection to a Time Server I would like to avoid, because we are often out of any Internet range)


ronzeiller:
@sarfata 
Just recognised that 
`const SKUpdate& SKNMEA2000Parser::parse(const SKSourceInput& input, const tN2kMsg& msg, const SKTime& timestamp) {`
in SKNMEA2000Parser.cpp is never entered.
Therefore no parsing even though N2k messages are coming in.

Any ideas?

sarfata:
Answering your questions:

> 1.) I am not sure if I can use for new PGNs: 
> `SKUpdateStatic<2> *update = new SKUpdateStatic<2>();`
> or need 
> `SKUpdateStatic<3> ...`
> or something else....

The number is the maximum number of value you are going to add. If you set 2 but you add 3 variables, the last one will be ignored.
 
> 2.) changed if-statement  to switch as this should reduce CPU usage?

In general I trust the compiler to do a better job optimizing these type of things than I could. Here I think the most important thing is readability of the code. I think the `switch` statement is the right choice here!

:+1: 

> 3.) added defined offset if not set in the Water Depth PGN
> ```
> if (!N2kIsNA(offset)) {
>       // When no offset then offset should be Zero
>       offset = 0;
>     }
> ```

Yes! Thank you!

> 4.) prepared (commented out) for config-settings if someone doesn't need a PGN converted 
> As soon config or variables for that is more defined we could do something like this or  similar
> 
> ```
> //if ( cfEnableN2kToNMEA_127250 ) {
>    return parse127250(input, msg, timestamp);
> //}
> ```

I don't think we should commit commented code anywhere (and yes I am guilty of doing it sometimes but I try not to!).
I think it will be just as easy to add the `if` statements when we have a real configuration system. Until then, those lines are just noise that are going to confuse someone who will read this. They might wonder why it's here, what was the intention. So here I think we should remove them.

> 5.) The parsing of Date & Time PGN 126992 could also be used for setting global true KBox System Date or Time.
> I had this already working in on of my KBox branches ("old" master), which was also writing date/time for logging files on SD and automatically switched to a new logfile ad midnight.
>
> I could start to implement this here by extending your SKTime class and add some (global) static variables. Or do you have a better idea may be?

Why do you think you need a global variable? We could do something like this: https://github.com/PaulStoffregen/Time/blob/master/examples/TimeTeensy3/TimeTeensy3.ino#L27

> Another way for getting true Date/Time is from NMEA GPS  (may be selecting this by config?)
> (Should be selectable, as some users take KBox without any NMEA input and others wants it without NMEA2000 at all.....)

I think we can support parsing both for now and in the future have a configuration option to ignore time coming from one specific channel. Most users will only have one time source and that will work fine.

> (A WiFi connection to a Time Server I would like to avoid, because we are often out of any Internet range)

src/common/signalk/signalk.json
```diff
@@ -36,6 +36,54 @@
       "units": "K",
       "description": "Current water temperature"
     },
+    {
+      "path": "environment.wind",
```
sarfata: This is not actually a value in SignalK. You cannot just provide `environment.wind` so we do not need this one.

src/common/signalk/signalk.json
```diff
@@ -36,6 +36,54 @@
...
+      "description": "Wind Data"
+    },
+    {
+      "path": "environment.wind.angleApparent",
+      "type": "numberValue",
+      "units": "rad",
+      "description": "AWA, Apparent Wind Angle"
+    },
+    {
+      "path": "environment.wind.angleTrue",
```
sarfata: `.angleTrue` [does not exist in SignalK](https://github.com/SignalK/specification/blob/master/schemas/groups/environment.json#L304). They have `angleTrueGround` and `angleTrueWater` which I think is good because it makes the difference between true wind calculated from the boat speed or true wind calculated from ground speed.

For speed they have `speedTrue` and `speedOverGround`. The descriptions are explicit about how it should be calculated.

For the description I just copy/paste the same description that is found in SignalK.

I had already added a large batch of wind variables that I just merged in `develop` which is why we now have a merge conflict. I am going to clean that up to make it easy for you to merge.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -43,19 +46,170 @@ const SKUpdate& SKNMEA2000Parser::parse(const SKSourceInput& input, const tN2kMs
...
+    default:
+      DEBUG("No known conversion for PGN %i", msg.PGN);
+      return _invalidSku;
+  } // end switch msg.PGN
+}
+
+// System Time / Date
+// TODO: Set system date and time for SD-Card logging file date and
+// Offset to millis() for recalculation of true date/time
+// TODO: after 42 days millis() overflow, new time sync needed
```
sarfata: KBox has a real-time clock: https://www.pjrc.com/teensy/td_libs_Time.html#teensy3. This works on your teensy 3.6 too. Just need to add a crystal on the board.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -43,19 +46,170 @@ const SKUpdate& SKNMEA2000Parser::parse(const SKSourceInput& input, const tN2kMs
...
+    update->setSteeringRudderAngle(RudderPosition);
+    DEBUG("Rudder: %d",SKRadToDeg(RudderPosition));
+    _sku = update;
+    return *_sku;
   }
-  // Water Depth
-  if (msg.PGN == 128267) {
-    return parse128267(input, msg, timestamp);
+  else {
+    INFO("Unable to parse NMEA2000 with PGN %i", msg.PGN);
```
sarfata: I think this should be a `DEBUG` because it could get repeated a lot if you are on a network where something is sending a PGN we do not understand. 

The theory is that INFO message should be non-verbose enough that they can be kept on all the time and even logged to a file.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -43,19 +46,170 @@ const SKUpdate& SKNMEA2000Parser::parse(const SKSourceInput& input, const tN2kMs
...
+  // nothing todo at the moment....
+  return _invalidSku;
+}
+
+// *****************************************************************************
+//   PGN 127250 VESSEL HEADING RAPID
+// *****************************************************************************
+const SKUpdate& SKNMEA2000Parser::parse127250(const SKSourceInput& input, const tN2kMsg& msg, const SKTime& timestamp) {
+  unsigned char sid;
+  tN2kHeadingReference HeadingReference;
```
sarfata: In KBox the coding style follows Google coding style.

All variables that are not `const` are in lower case so here it should be `headingReference`, `heading`, `deviation`, `variation`.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -43,19 +46,170 @@ const SKUpdate& SKNMEA2000Parser::parse(const SKSourceInput& input, const tN2kMs
...
+  if (ParseN2kHeading(msg,sid,Heading,Deviation,Variation,HeadingReference) ) {
+
+    if ( String(HeadingReference) == "magnetic" ) {
+      // Deviation and Variation could be sent
+      // (Timo's NMEA Simulator is sending Dev and Var at magnetic HDG only)
+      Heading += Deviation + Variation;
+    }
+    // if ( String(HeadingReference) == "true" ) {  }
+
+    SKUpdateStatic<2> *update = new SKUpdateStatic<2>();
```
sarfata: This causes a memory leak because you are creating a new `SKUpdateStatic` but you are not keeping the pointer. You should assign it to `_sku` immediately to make sure we do not "lose" it.

If you assign it to `_sku` the `parse()` method will delete it next time a NMEA2000 message is received.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -43,19 +46,170 @@ const SKUpdate& SKNMEA2000Parser::parse(const SKSourceInput& input, const tN2kMs
...
+}
+
+// *****************************************************************************
+//   PGN 127250 VESSEL HEADING RAPID
+// *****************************************************************************
+const SKUpdate& SKNMEA2000Parser::parse127250(const SKSourceInput& input, const tN2kMsg& msg, const SKTime& timestamp) {
+  unsigned char sid;
+  tN2kHeadingReference HeadingReference;
+  double Heading = 361; // not valid value
+  double Deviation = 0; // NAN would be better!
```
sarfata: You do not need to initialize these variables. The NMEA2000 library [always sets a value for them](https://github.com/ttlappalainen/NMEA2000/blob/master/src/N2kMessages.cpp#L100).

If you want to initialize them (which is never a bad idea, +1 for good coding practices ;) ) then you can use `N2kDoubleNA`.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -43,19 +46,170 @@ const SKUpdate& SKNMEA2000Parser::parse(const SKSourceInput& input, const tN2kMs
...
+  double Variation = 0; // NAN would be better!
 
-  DEBUG("No known conversion for PGN %i", msg.PGN);
+  if (ParseN2kHeading(msg,sid,Heading,Deviation,Variation,HeadingReference) ) {
+
+    if ( String(HeadingReference) == "magnetic" ) {
```
sarfata: You can just do `headingReference == N2khr_magnetic` (it is defined [here](https://github.com/ttlappalainen/NMEA2000/blob/master/src/N2kMessages.h#L60)).

I think changing the `heading` value based on it's type is wrong. What we should do is set a different signalK variable.

If it's magnetic, we set: `headingMagnetic` and `magneticVariation` (if the variation is set). SignalK does not support deviation now but I think it should be added. I will ask them why it's not here now.

If it's true, we set `headingTrue`.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -27,19 +27,85 @@
   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
   THE SOFTWARE.
 */
-
+#include <Arduino.h>
```
sarfata: Why do you need `Arduino.h` here? It's a very big header that includes many other headers and it makes it hard to compile code on other platforms. 

I try to always be more specific about what I need and only include the specific header that I need.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -27,19 +27,85 @@
...
+//  8) K = Kilometers Per Hour
+// ************************************************************************
+void SKNMEAVisitor::visitSKEnvironmentWindAngleTrueWater(const SKUpdate &u, const SKPath &p, const SKValue &v) {
+
+  float WindAngle;
+  float WindSpeed;
+
+  WindAngle = SKRadToDeg(v.getNumberValue());
+
+  // WindAngle and WindSpeed should come together......
```
sarfata: Yes, so you can just check if the windSpeed is there or return which is what you are doing and that is good.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -27,19 +27,85 @@
...
+//  3) Speed
+//  4) N = Knots
+//  5) Speed
+//  6) M = Meters Per Second
+//  7) Speed
+//  8) K = Kilometers Per Hour
+// ************************************************************************
+void SKNMEAVisitor::visitSKEnvironmentWindAngleTrueWater(const SKUpdate &u, const SKPath &p, const SKValue &v) {
+
+  float WindAngle;
```
sarfata: Please use `windAngle`.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -27,19 +27,85 @@
...
+
+  // Validation
+  if (WindAngle<0||WindAngle>360||!u.hasEnvironmentWindSpeedTrue()) return;
+
+  WindSpeed = u.getEnvironmentWindSpeedTrue();  // in m/s
+  DEBUG("WindSpeed: %f", WindSpeed);
+
+  NMEASentenceBuilder sb( "II", "MWV", 5);
+  sb.setField(1, WindAngle, 1 );
+  sb.setField(2, "R");
```
sarfata: This should be `T` because we are measuring true wind here.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -27,19 +27,85 @@
...
+  float WindAngle;
+  float WindSpeed;
+
+  WindAngle = SKRadToDeg(v.getNumberValue());
+
+  // WindAngle and WindSpeed should come together......
+  // TODO: one visit enough? which is coming first, Angle or Speed????
+
+  // Validation
+  if (WindAngle<0||WindAngle>360||!u.hasEnvironmentWindSpeedTrue()) return;
```
sarfata: Are you sure that in NMEA2000 the windAngle is going to be between 0 and 360? I would not be surprised if it was -180 to 180. Maybe we should be more flexible here.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -27,19 +27,85 @@
...
+  NMEASentenceBuilder sb( "II", "MWV", 5);
+  sb.setField(1, WindAngle, 1 );
+  sb.setField(2, "R");
+  sb.setField(3, WindSpeed, 1 );
+  sb.setField(4, "M");
+  sb.setField(5, "A");
   _sentences.add(sb.toNMEA() + "\r\n");
+  //Serial.printf("%s\n", sb.toNMEA());
+
+  NMEASentenceBuilder sb1( "II", "VWR", 8);
```
sarfata: I think we should only send this when dealing with apparentWind, not for true wind.

My "bible" does not specify this:
http://www.catb.org/gpsd/NMEA.html#_vwr_relative_wind_speed_and_angle

But since it's called "relative wind speed and angle", I think it implies that this is for apparent. No?

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -27,19 +27,85 @@
...
+    sb1.setField(2, "L");
+  }
+  sb1.setField(3, SKMsToKnots( WindSpeed ), 2 );
+  sb1.setField(4, "N");
+  sb1.setField(5, WindSpeed, 2);
+  sb1.setField(6, "M");
+  sb1.setField(7, SKMsToKmh( WindSpeed ), 2);
+  sb1.setField(8, "K");
+  _sentences.add(sb1.toNMEA() + "\r\n");
+  //Serial.printf("%s\n", sb.toNMEA());
```
sarfata: Please remove old comments :)

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -51,6 +117,17 @@ void SKNMEAVisitor::visitSKEnvironmentOutsidePressure(const SKUpdate& u, const S
   sb.setField(3, "B");
   sb.setField(4, "Barometer");
   _sentences.add(sb.toNMEA() + "\r\n");
+  //Serial.printf("%s\n", sb.toNMEA());
+}
+
+void SKNMEAVisitor::visitSKElectricalBatteriesVoltage(const SKUpdate& u, const SKPath &p, const SKValue &v) {
```
sarfata: Have you tested this with some tools? (I am just curious - any output is better than no output).

src/common/signalk/SKUnits.h
```diff
@@ -32,3 +32,6 @@
 
 #define SKDegToRad(x) x * 2 * M_PI / 360
 #define SKRadToDeg(x) x * 360.0 / (2 * M_PI)
+
+#define SKMsToKnots(x) x * 3600.0 / 1852.0
```
sarfata: Just to be consistent with the other conversions I think we should say `SKMsToKnot()` here (instead of knot**s**).

src/host/services/NMEA2000Service.cpp
```diff
@@ -39,7 +39,7 @@
 static NMEA2000Service *handlerContext;
 
 static void handler(const tN2kMsg &msg) {
-  DEBUG("Received N2K Message with pgn: %i", msg.PGN);
+  //DEBUG("Received N2K Message with pgn: %i", msg.PGN);
```
sarfata: I agree this message is annoying and we should have a way to disable it. I think commenting it is ok until we have a better way.

src/host/services/NMEA2000Service.cpp
```diff
@@ -61,14 +61,14 @@ void NMEA2000Service::setup() {
...
+  /*
   DEBUG("Sending message on n2k bus - pgn=%i prio=%i src=%i dst=%i len=%i result=%s", msg.PGN, msg.Priority,
       msg.Source,
       msg.Destination, msg.DataLen, result ? "success":"fail");
-
-  char pcdin[100];
-  N2kToSeasmart(msg, now(), pcdin, sizeof(pcdin));
-  DEBUG("TX: %s", pcdin);
+  */
+  //char pcdin[100];
```
sarfata: This is an example of a message that should not be here anymore. Please remove it completely. Sorry about that!

sarfata:
This is an awesome contribution with logs of very useful parsing. The biggest thing it is missing are tests. Can you please make sure to add tests in `SKNMEAVisitorTest.cpp` and `SKNMEA2000ParserTest.cpp`? It should not take long and it will help us make sure that we do not break all of this in the future.

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: I will investigate this.
I think a battery for keeping time is not so good (as it is not in KBox too) and the power for navigation is turned off in the marina.
GPS from NMEA and/or N2k is fine.

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: I'll do all you mentioned here and above.
Put the code for deviation commented to be ready when Signal K is.
(I know, that I should not put commented code... :smirk:)

Could you please ask them for support of reference type of speed through water too, as we will have to correct readings from paddle wheel by the cosinus of leeway later on.
(Ultra sound and doppler sound do not need this correction)
**Reference types:**
//                Paddle_wheel=0,
//                Pitot_tube=1,
//                Doppler_log=2,
//                Ultra_Sound=3,
//                Electro_magnetic=4

ronzeiller: I'll do all you mentioned here and above.
Put the code for deviation commented to be ready when Signal K is.
(I know, that I should not put commented code... :smirk:)

Could you please ask them for support of reference type of speed through water too, as we will have to correct readings from paddle wheel by the cosinus of leeway later on.
(Ultra sound and doppler sound do not need this correction)
**Reference types:**
//                Paddle_wheel=0,
//                Pitot_tube=1,
//                Doppler_log=2,
//                Ultra_Sound=3,
//                Electro_magnetic=4

src/common/signalk/SKUnits.h
ronzeiller: yes of course, you are right!

src/common/signalk/SKNMEAVisitor.cpp
ronzeiller: I could not get the BMP280 working here in I2C mode :-( so I do not know anything about pressure.
Voltage is quite ok. But could get an offset due to different electronics. 
Or just different settings in board.h

src/common/signalk/SKNMEAVisitor.cpp
ronzeiller: Sorry for that.
There is a lot of rubbish in the code.
I thought it would be an idea to send you something to discuss details in a comfortable way, but.....

src/common/signalk/SKNMEAVisitor.cpp
ronzeiller: no, definitely not! I mixed up different NMEA sentences.
In the new version I do not code the deprecated sentence anymore
You probably know this anyway:
https://www.nmea.org/Assets/100108_nmea_0183_sentences_not_recommended_for_new_designs.pdf

src/common/signalk/SKNMEA2000Parser.cpp
sarfata: The RTC is independent of a battery. The battery just means that the time is kept when you power off.

With or without a battery, the RTC is a better way to measure time than the `millis()` function. It will be much more precise (because it uses a 32khz quartz) and it is simpler for us because the hardware does it automatically.

src/common/signalk/SKNMEA2000Parser.cpp
sarfata: I have asked in the #specification channel on the SignalK slack group. You should join. Unless they have a reason to not include it, I guess the best thing to do would be to submit a pull-request to them ;)

src/common/signalk/SKNMEAVisitor.cpp
sarfata: It's totally fine to have it when we discuss ;) I just wanted to take a deep look and point to those things so we agree they need to be removed. I will not do that in the future if you say it's a draft ;)

sarfata:
I have taken a quick look at your other branch but you can either update this branch with your new code when you are done. Or close this PR and open a new one with the other branch.

Thanks again!

ronzeiller:
Hi Thomas,
I feel that I am doing something wrong...
More Parsing of N2k sentences and more building of NMEA0183 sentences is ok, but how to get them working?
At the moment nothing calls the 
```
SKNMEA2000Parser::parse(const SKSourceInput& input,....
```

OK, I could do something like:

```
SKNMEA2000Parser nmea2000Parser = SKNMEA2000Parser();
nmea2000Parser.parse(SKSourceInputNMEA2000, msg, n2km.getReceivedTime());
```

in KMessageNMEAVisitor.cpp 
but this is definitely not what you are going for....

Also, there is nothing to get the new NMEA0183 sentences when an according PGN comes from the N2k-Bus.

I thought somehow the visitor will solve it, but it was naive thinking.

At the moment I am quite lost ..... :confused:

sarfata:
It’s not your fault. I have not plugged in the nmea2000visitor yet. I have
just used the tests to verify it’s working.

You do the same for now. I will make a PR to actually send the messages.
Will be pretty easy but I have a few things I might improve along the way.

If you want to do it yourself the NMEA2000Service needs to register to
receive updates from the SKHub and process them. See line 43 and 83 of the
WiFiService for an example:
https://github.com/sarfata/kbox-firmware/blob/develop/src/host/services/WiFiService.cpp#L83

Thomas

ronzeiller:
OK, then I will let you do this, because you know what you do :grin:
And I close this PR now, because you got a new one.

PR #84, opened 2017-11-28
more parsing

ronzeiller:
OK, here comes the new "more parsing" with 

- new Wind, Rudder, Vessel Heading rapid (true & magnetic) -> taking heading from N2k bus or internal sensor should become a setting
- Tests added in SKNMEA2000ParserTest
- small correction function name toNMEA -> getNMEAContent() (as mentioned in your comments)


src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -42,20 +43,65 @@ const SKUpdate& SKNMEA2000Parser::parse(const SKSourceInput& input, const tN2kMs
   if (_sku) {
     delete(_sku);
   }
+  //RONDEBUG
```
sarfata: Did you mean to keep this ;)?

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -99,23 +149,155 @@ const SKUpdate& SKNMEA2000Parser::parse128267(const SKSourceInput& input, const
...
-      }
-      else if (offset > 0) {
-        update->setEnvironmentDepthSurfaceToTransducer(offset);
-        update->setEnvironmentDepthBelowSurface(depthBelowTransducer + offset);
+    if (N2kIsNA(offset)) {
+      // When no offset then offset should be Zero
+      offset = 0;
+    }
+    // When offset is negative, it's the distance between transducer and keel
+    if (offset < 0) {
```
sarfata: Am I missing something? Or is this is not going to work when the offset is 0. I am surprised the tests did not catch this, we should have a test for this.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -99,23 +149,155 @@ const SKUpdate& SKNMEA2000Parser::parse128267(const SKSourceInput& input, const
...
+    return _invalidSku;
+  }
+}
+
+// *****************************************************************************
+//   PGN 130306 W I N D
+//    tN2kWindReference:
+//      N2kWind_True_North=0,
+//      N2kWind_Magnetic=1,
+//      N2kWind_Apprent=2,
```
sarfata: Looks like a typo in NMEA2000. @ttlappalainen is this supposed to be `Apprent`? I think `apparent` would be better.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -99,23 +149,155 @@ const SKUpdate& SKNMEA2000Parser::parse128267(const SKSourceInput& input, const
...
+    update->setTimestamp(timestamp);
+
+    SKSource source = SKSource::sourceForNMEA2000(input, msg.PGN, msg.Priority, msg.Source);
+    update->setSource(source);
+
+    if (!N2kIsNA( windAngle) && !N2kIsNA( windSpeed)) {
+      if (windReference == N2kWind_Apprent) {
+        // -> m/s (Meters per second)
+        update->setEnvironmentWindSpeedApparent(SKKnotToMs(windSpeed));
+        DEBUG("AWS: %d kts",windSpeed);
```
sarfata: please remove extra log messages ;)

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -99,23 +149,155 @@ const SKUpdate& SKNMEA2000Parser::parse128267(const SKSourceInput& input, const
...
+        // -> Apparent wind angle, rad, negative to port
+        if (windAngle >M_PI) windAngle *= -1;
+        update->setEnvironmentWindAngleApparent(windAngle);
+        DEBUG("AWA: %d °",SKRadToDeg(windSpeed));
       }
     }
+    // We do not take True Wind Speed as we do not know how the system has calculated it!!
```
sarfata: I think we do. Look at the different options: https://github.com/ttlappalainen/NMEA2000/blob/master/src/N2kMessages.h#L150

And then we can also take inspiration from [what n2k-signalk does with this PGN](https://github.com/SignalK/n2k-signalk/blob/master/pgns/130306.js):
- Apparent is obvious => `environment.wind.speedApparent` and `environment.wind.angleApparent`.
- True boat is true calculated to the boat speed => `environment.wind.speedTrue` and `environment.wind.angleTrueWater`
- True water is true calculated to the SOG => `environment.wind.speedOverGround` and `environment.wind.directionTrue`
- The last two ones are harder. `n2k-signalk` does not support them. I would guess that they are both apparent but using the boat heading to calculate a wind direction instead of a wind angle. With true north, the angle has just been corrected for the declination. => `environment.wind.speedApparent` and `environment.wind. directionTrue` or `.directionMagnetic` (I am not sure about this one).

@ttlappalainen do you have any tips for us here?

I think the best thing to do here would be to collect some samples from boats (do you have a nmea2000 sensor?) and add some tests to make sure we are doing it right.

You have found the unit tests but did you look [at the integration tests](https://github.com/sarfata/kbox-firmware/blob/develop/src/sktool/tests/n2k-transducer.json)? I have not documented this at all but it's fun to try! Travis runs them automatically. On Linux/Mac you can try with just `make validation` or just run `src/sktool/sktool-test.py --implementation sktool src/sktool/tests/*json` to run them locally.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -99,23 +149,155 @@ const SKUpdate& SKNMEA2000Parser::parse128267(const SKSourceInput& input, const
...
+
+    if (!N2kIsNA( windAngle) && !N2kIsNA( windSpeed)) {
+      if (windReference == N2kWind_Apprent) {
+        // -> m/s (Meters per second)
+        update->setEnvironmentWindSpeedApparent(SKKnotToMs(windSpeed));
+        DEBUG("AWS: %d kts",windSpeed);
+        // -> Apparent wind angle, rad, negative to port
+        if (windAngle >M_PI) windAngle *= -1;
+        update->setEnvironmentWindAngleApparent(windAngle);
+        DEBUG("AWA: %d °",SKRadToDeg(windSpeed));
```
sarfata: This debug is wrong. Converts a speed into an angle.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -99,23 +149,155 @@ const SKUpdate& SKNMEA2000Parser::parse128267(const SKSourceInput& input, const
...
+    SKUpdateStatic<2> *update = new SKUpdateStatic<2>();
+    update->setTimestamp(timestamp);
+
+    SKSource source = SKSource::sourceForNMEA2000(input, msg.PGN, msg.Priority, msg.Source);
+    update->setSource(source);
+
+    if (!N2kIsNA( windAngle) && !N2kIsNA( windSpeed)) {
+      if (windReference == N2kWind_Apprent) {
+        // -> m/s (Meters per second)
+        update->setEnvironmentWindSpeedApparent(SKKnotToMs(windSpeed));
```
sarfata: This is incorrect. The NMEA2000 source code indicates that this should already be in m/s. See https://github.com/ttlappalainen/NMEA2000/blob/master/src/N2kMessages.h#L1204.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -99,23 +149,155 @@ const SKUpdate& SKNMEA2000Parser::parse128267(const SKSourceInput& input, const
...
+  double heading = N2kDoubleNA;
+  double deviation = N2kDoubleNA;
+  double variation = N2kDoubleNA;
+
+  if (ParseN2kHeading(msg,sid,heading,deviation,variation,headingReference) ) {
+
+    if (headingReference == N2khr_magnetic) {
+      // Deviation and Variation could be sent
+      // (Timo's NMEA Simulator is sending Dev and Var at magnetic HDG only)
+      heading += deviation + variation;
```
sarfata: I do not think we should transform the heading. If it's magnetic we should just pass the fact that it is magnetic but not transform it.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -99,23 +149,155 @@ const SKUpdate& SKNMEA2000Parser::parse128267(const SKSourceInput& input, const
...
+      if (!N2kIsNA(variation))
+        update->setNavigationMagneticVariation(variation);
+        /* coming when Signal K adds deviation
+        if (!N2kIsNA(deviation))
+          update->setNavigationMagneticDeviation(deviation);
+        */
+      _sku = update;
+      return *_sku;
+    }
+    if ( headingReference == N2khr_true ) {
```
sarfata: (nitpicking) why do you have extra spaces here around the expression? You do not have them above.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -99,23 +149,155 @@ const SKUpdate& SKNMEA2000Parser::parse128267(const SKSourceInput& input, const
...
+      SKUpdateStatic<1> *update = new SKUpdateStatic<1>();
+      update->setTimestamp(timestamp);
+
+      SKSource source = SKSource::sourceForNMEA2000(input, msg.PGN, msg.Priority, msg.Source);
+      update->setSource(source);
+      update->setNavigationHeadingTrue(heading);
```
sarfata: Are you sure that headingTrue never includes information on the deviation? Would be useful for a client to calculate the magnetic heading. I think we could support this with very little extra effort.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -33,13 +33,17 @@
...
+#include <KBoxLogging.h>
+
 void SKNMEAVisitor::visitSKElectricalBatteriesVoltage(const SKUpdate& u, const SKPath &p, const SKValue &v) {
   NMEASentenceBuilder sb("II", "XDR", 4);
   sb.setField(1, "V");
   sb.setField(2, v.getNumberValue(), 2);
   sb.setField(3, "V");
   sb.setField(4, p.getIndex());
   _sentences.add(sb.toNMEA() + "\r\n");
+  DEBUG( sb.toNMEA().c_str() );
```
sarfata: Please remove this. If you want to test this, you should just write a test case. It's a much better and faster way to debug these the converters, visitors, etc.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -68,12 +73,98 @@ void SKNMEAVisitor::visitSKNavigationAttitude(const SKUpdate &u, const SKPath &p
...
+//  5) Speed
+//  6) M = Meters Per Second
+//  7) Speed
+//  8) K = Kilometers Per Hour
+// ************************************************************************
+void SKNMEAVisitor::visitSKEnvironmentWindAngleApparent(const SKUpdate &u, const SKPath &p, const SKValue &v) {
+
+  float windAngle;
+  float windSpeed;
+  windAngle = SKRadToDeg(v.getNumberValue());
```
sarfata: I think we should do the math in radians here. Only convert at the end when we print.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -68,12 +73,98 @@ void SKNMEAVisitor::visitSKNavigationAttitude(const SKUpdate &u, const SKPath &p
...
+
+  // Validation
+  if (windAngle>180||!u.hasEnvironmentWindSpeedApparent()) return;
+
+  windSpeed = u.getEnvironmentWindSpeedTrue();  // in m/s
+  DEBUG("WindSpeed: %f", windSpeed);
+
+  NMEASentenceBuilder sb( "II", "VWR", 8);
```
sarfata: In your other PR you said you would [remove VWR because it is not recommended for new design](https://www.nmea.org/Assets/100108_nmea_0183_sentences_not_recommended_for_new_designs.pdf) and use MWV instead. It looks like what you did is the opposite of what you wanted to do.

src/test/signalk/SKNMEA2000ParserTest.cpp
```diff
@@ -83,5 +83,35 @@ TEST_CASE("SKNMEA2000Parser: Basic tests") {
     CHECK( update.getEnvironmentDepthBelowKeel() == 40.8 );
     CHECK( update.getEnvironmentDepthTransducerToKeel() == 1.2 );
   }
-}
 
+  SECTION("127250: VESSEL True HEADING RAPID ") {
+    SetN2kTrueHeading(msg, 0, 180);
+    const SKUpdate &update = p.parse(SKSourceInputNMEA2000, msg, SKTime(0));
+    CHECK( update.getSize() == 1);
+    CHECK( update.getNavigationHeadingTrue() == 180 );
```
sarfata: This test does not work. See Travis output: https://travis-ci.org/sarfata/kbox-firmware/builds/308718735#L1692

You need to do ` == SKDegToRad(180)` here.

src/test/signalk/SKNMEA2000ParserTest.cpp
```diff
@@ -83,5 +83,35 @@ TEST_CASE("SKNMEA2000Parser: Basic tests") {
...
+    CHECK( update.getSize() == 2);
+    CHECK( update.getNavigationHeadingMagnetic() == 181 );
+    // CHECK( update.getNavigationMagneticDeviation() == -2 );
+    CHECK( update.getNavigationMagneticVariation() == 3 );
+  }
+
+  SECTION("127245: Rudder Angle") {
+    SetN2kRudder(msg,-3.5, 0);
+    const SKUpdate &update = p.parse(SKSourceInputNMEA2000, msg, SKTime(0));
+    CHECK( update.getSize() == 1);
```
sarfata: For some reason this test failed too: https://travis-ci.org/sarfata/kbox-firmware/builds/308718735#L1716

Apparently the message was not processed at all (test output says `update.getSize() == 0`).

src/test/signalk/SKNMEA2000ParserTest.cpp
```diff
@@ -83,5 +83,35 @@ TEST_CASE("SKNMEA2000Parser: Basic tests") {
...
+  SECTION("127250: VESSEL Magnetic HEADING RAPID ") {
+    SetN2kMagneticHeading(msg, 0, 181, -2, 3);
+    const SKUpdate &update = p.parse(SKSourceInputNMEA2000, msg, SKTime(0));
+    CHECK( update.getSize() == 2);
+    CHECK( update.getNavigationHeadingMagnetic() == 181 );
+    // CHECK( update.getNavigationMagneticDeviation() == -2 );
+    CHECK( update.getNavigationMagneticVariation() == 3 );
+  }
+
+  SECTION("127245: Rudder Angle") {
```
sarfata: We should also have a test with a rudder direction order. Even if we do not support it, it will make it explicit that we don't.

src/test/signalk/SKNMEA2000ParserTest.cpp
```diff
@@ -83,5 +83,35 @@ TEST_CASE("SKNMEA2000Parser: Basic tests") {
...
+    SetN2kRudder(msg,-3.5, 0);
```
sarfata: You need to convert in radians here. Otherwise -3.5 is not what you think it is.

src/test/signalk/SKNMEA2000ParserTest.cpp
```diff
@@ -83,5 +83,35 @@ TEST_CASE("SKNMEA2000Parser: Basic tests") {
...
+    const SKUpdate &update = p.parse(SKSourceInputNMEA2000, msg, SKTime(0));
+    CHECK( update.getSize() == 1);
+    CHECK( update.getSteeringRudderAngle() == -3.5 );
```
sarfata: and here convert to degrees.

src/test/signalk/SKNMEA2000ParserTest.cpp
```diff
@@ -83,5 +83,35 @@ TEST_CASE("SKNMEA2000Parser: Basic tests") {
     CHECK( update.getEnvironmentDepthBelowKeel() == 40.8 );
     CHECK( update.getEnvironmentDepthTransducerToKeel() == 1.2 );
   }
-}
 
+  SECTION("127250: VESSEL True HEADING RAPID ") {
+    SetN2kTrueHeading(msg, 0, 180);
```
sarfata: The value 180 is in radians here. So it's not what you think it is ;) Use `SKDegToRad(180)`

src/test/signalk/SKNMEA2000ParserTest.cpp
```diff
@@ -83,5 +83,35 @@ TEST_CASE("SKNMEA2000Parser: Basic tests") {
...
+  }
+
+  SECTION("127245: Rudder Angle") {
+    SetN2kRudder(msg,-3.5, 0);
+    const SKUpdate &update = p.parse(SKSourceInputNMEA2000, msg, SKTime(0));
+    CHECK( update.getSize() == 1);
+    CHECK( update.getSteeringRudderAngle() == -3.5 );
+  }
+
+  SECTION("130306: W I N D") {
```
sarfata: Same comment about the angles. And we should add a test for true boat and true ground types too.

sarfata:
One more thing, please do not create a new PR in the future. Continue in this one with your branch. This way we keep the history of the conversation.

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: I should not make PRs late night....
Sorry for all those careless mistakes.

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: I should have written "True Wind will be done later..." :smiley:

1. I think it will be of importance, that a user can choose if she wants TrueWind from the bus or calculated in KBox 

2. True Boat makes sense as it should be the "True (Sailing) Wind" referred to water

3. N2kWind_True_North=0 ->  is true calculated to the SOG => environment.wind.speedOverGround and environment.wind.directionTrue 

=> I will follow up with a translation of 2. and 3.

4. N2kWind_Magnetic=1 => further investigations needed

To your Questions:
I built a simple data-logger with a modified KBox-program, but had no chance to be on a boat yet.

It would be nice to have more of different logfiles @ttlappalainen mentioned once that he has a lot of them.
My only source at the moment is:
Kees Merrimac file: https://github.com/canboat/canboat/blob/master/samples/merrimac-actisense-serial-2011.raw
(I just saw that there are some new files there, I will have to look at)

More of those files would give us a nice opportunity to do some "live"-tests

I found the other test files, but could not sort it out, how to handle them. 
Thank you for your description, looking forward to try it!

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: OMG a bad BUG!!
Of course not!!!
(....now I know, what you meant in the discussion of the last PR....)

ronzeiller: I did an additionally validation check:

`if (!N2kIsNA(heading) && heading >= 0 && heading <= M_PI_2) {`

What do you think?

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: :smiley:
corrected all I found

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: I think I remember, that I have looked at Timos NMEA Simulator to find out that it is just sending with the Mag Hdg.

Pros and Cons:
Yes, it could/should be done with a config-setting if someone wants it or not.
No, IMHO Heading and Depth have to be treated very carefully and should never be changed or calculated if the user doesn't know it.

src/common/signalk/SKNMEAVisitor.cpp
ronzeiller: Yes, agreed.
I had it in the code to do the maths only once, but now, I cancelled the deprecated NMEA0183 sentence.
(To be honest, I do not know what is heavier for CPU doing calculations twice or invent a new variable and do it once)

src/common/signalk/SKNMEAVisitor.cpp
ronzeiller: Yes, you are right!!

src/test/signalk/SKNMEA2000ParserTest.cpp
ronzeiller: now I (hope to) know more about this tests I will do them before doing a PR
In this case it could be the angle too (which should be rad again)

src/test/signalk/SKNMEA2000ParserTest.cpp
ronzeiller: will come with the parsing of discussed other two wind types.

ronzeiller:
The failures:
```
src/test/signalk/SKNMEA2000ParserTest.cpp:65: FAILED:
  CHECK( update.getSize() == 1 )
with expansion:
  3 == 1
```
And at Rudder, I have no idea why getSize returns 0 (and value is not stored then?)
```
src/test/signalk/SKNMEA2000ParserTest.cpp:106: FAILED:
  CHECK( update.getSize() == 1 )
with expansion:
  0 == 1
src/test/signalk/SKNMEA2000ParserTest.cpp:107: FAILED:
  CHECK( update.getSteeringRudderAngle() == -0.0157 )
with expansion:
  0.0 == -0.0157
```
this needs some further investigation.

Generally may be something like would be better?:
`CHECK( (update.getEnvironmentWindAngleApparent() - SKDegToRad(29.8) ) < 0.01  );`

Last test failed e.g.:
```
src/test/signalk/SKNMEA2000ParserTest.cpp:98: FAILED:
  CHECK( update.getNavigationHeadingMagnetic() == 0.7898 )
with expansion:
  0.7898 == 0.7898
```

Makes it easier, as not always knowing how many decimals are stored in N2kmsg
(I will give it a try...)

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: But, we could also translate PGN127258 Magnetic Variation which might come from GPS/Plotter

ronzeiller:
By the way:
How an addition to the signalk.json should be made for:

**/vessels/<RegExp>/environment/current**

Title: current
Description: Direction and strength of current affecting the vessel

**Object value with properties**
drift (m/s)
setTrue (rad)
setMagnetic (rad)

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -80,42 +128,207 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
+        // Ground Wind Direction
+        update->setEnvironmentWindDirectionTrue(windAngle);
+      }
+      else if (windReference == N2kWind_Magnetic) {
+        // Ground Wind Speed
+        update->setEnvironmentWindSpeedOverGround(windSpeed);
+        // Ground Wind Direction refered to magnetic north
+        update->setEnvironmentWindDirectionMagnetic(windAngle);
+      }
+      //TODO: correct if Timos Typo will be corrected in library
```
sarfata: Will be fixed very soon https://github.com/ttlappalainen/NMEA2000/pull/85

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -80,42 +128,207 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
+      else if (windReference == N2kWind_Apprent) {
+        // AWS Apparent Wind Speed
+        update->setEnvironmentWindSpeedApparent(windSpeed);
+        // AWA pos coming from starboard, neg from port, relative to centerline vessel
+        if (windAngle > M_PI) windAngle *= -1;
```
sarfata: At first I thought this was ok but now I do not think this works. If you think about it in degrees, if the angle is 185 (slightly more than PI), then -1*185 will be -185 degrees which is 175 degrees. You have actually changed the value. Not good.

Maybe we should have a function (or a macro) to normalize an angle between -M_PI and +M_PI. 
Something like `#define NORMALIZE_ANGLE(x) fmod(x, 2*M_PI)-M_PI`. This way we do not risk repeating the same error and it will make the code more readable.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -80,42 +128,207 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
+//        →  RudderPosition [rad]
+// *  ********************************************** */
+const SKUpdate& SKNMEA2000Parser::parse127245(const SKSourceInput& input, const tN2kMsg& msg, const SKTime& timestamp) {
+  unsigned char instance;
+  tN2kRudderDirectionOrder rudderDirectionOrder;
+  double rudderPosition = N2kDoubleNA;
+  double angleOrder = N2kDoubleNA;
+
+  if (ParseN2kRudder(msg,rudderPosition,instance,rudderDirectionOrder,angleOrder)) {
+    // validation check max +/- 45°
```
sarfata: are you sure we need this? my rudder turns more than 45°. I was told by my surveyor that it is bad but that is what it is now.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -80,42 +128,207 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
+      update->setSteeringRudderAngle(rudderPosition);
+
+      _sku = update;
+      return *_sku;
+    }
+  }
+  DEBUG("Unable to parse NMEA2000 with PGN %i", msg.PGN);
+  return _invalidSku;
+}
+// *****************************************************************************
```
sarfata: very very small nitpick: can you include a blank line between functions?

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -80,42 +128,207 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
+const SKUpdate& SKNMEA2000Parser::parse127250(const SKSourceInput& input, const tN2kMsg& msg, const SKTime& timestamp) {
+  unsigned char sid;
+  tN2kHeadingReference headingReference;
+  double heading = N2kDoubleNA;
+  double deviation = N2kDoubleNA;
+  double variation = N2kDoubleNA;
+
+  if (ParseN2kHeading(msg,sid,heading,deviation,variation,headingReference)) {
+    if (!N2kIsNA(heading) && heading >= 0 && heading <= 2 * M_PI) {
```
sarfata: Are you sure we need to check that heading is between 0 and 2 PI? Other values are valid too but we could just normalize it to make sure whatever we publish is between 0 and 2 pi.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -80,42 +128,207 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
+  }
+  DEBUG("Unable to parse NMEA2000 with PGN %i", msg.PGN);
+  return _invalidSku;
+}
+
+// *****************************************************************************
+//    PGN 128000 Nautical Leeway Angle (new 2017)
+// https://www.nmea.org/Assets/20170204%20nmea%202000%20leeway%20pgn%20final.pdf
+// Upcoming in Timos library:
+// void SetN2kPGN128000(tN2kMsg &N2kMsg, unsigned char SID, double Leeway) {
```
sarfata: Timo has merged this now if you want (or it can stay as a comment and wait for a future PR).

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -71,9 +73,90 @@ void SKNMEAVisitor::visitSKNavigationAttitude(const SKUpdate &u, const SKPath &p
...
+//  8) K = Kilometers Per Hour
+// ************************************************************************
+void SKNMEAVisitor::visitSKEnvironmentWindAngleApparent(const SKUpdate &u, const SKPath &p, const SKValue &v) {
+
+  float windAngle;
+  float windSpeed;
+  windAngle = v.getNumberValue();
+
+  // Validation
+  if ((windAngle < 0)||(windAngle > 2 * M_PI)||(!u.hasEnvironmentWindSpeedApparent())) return;
```
sarfata: No. The values we generate when parsing NMEA2000 data will have an angle between -PI and +PI if it's an angle and not a direction. But here we would refuse this as an input and not generate an NMEA sentence for it.

sarfata: Why are you validating that we `hasEnvironmentWindSpeedApparent()` here but not on the other ones? I think we should do the same thing everywhere. Here it's a private method called by another method who does the `if()` for us so I think we can skip it.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -80,42 +128,207 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
+// *  ********************************************** */
+const SKUpdate& SKNMEA2000Parser::parse127245(const SKSourceInput& input, const tN2kMsg& msg, const SKTime& timestamp) {
+  unsigned char instance;
+  tN2kRudderDirectionOrder rudderDirectionOrder;
+  double rudderPosition = N2kDoubleNA;
+  double angleOrder = N2kDoubleNA;
+
+  if (ParseN2kRudder(msg,rudderPosition,instance,rudderDirectionOrder,angleOrder)) {
+    // validation check max +/- 45°
+    if (!N2kIsNA(rudderPosition) && (rudderPosition <= M_PI_4) && (rudderPosition >= M_PI_4) ) {
```
sarfata: Your condition says that `rudderPosition` must be less than 45° and more than 45°at the same time. That is not possible.

This is why the test does not pass (and why tests are so awesome!!!).

src/common/signalk/SKNMEAVisitor.cpp
ronzeiller: Good question!
Should we do some validations, or not?
of course they will have to work, not like mine.... ;-)

I was not clear here, this is why I sometimes do and sometimes not, more or less to discuss it.
May be just to check !N2kIsNA() or even nothing at all?
What is your opinion?

All the rest I will read carefully and do it later...

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: You are right of course!
The normalize macro is a great idea, where to put it best?
SKUnits.h ?

src/common/signalk/SKNMEAVisitor.cpp
ronzeiller: For the ´!u.hasEnvironmentWindSpeedApparent()´ validation:

**My logic was:**
For this sentence we need two different Signal K.
one is triggered by the visitor/update itself, here the angle.
The other one needed, I check.

**But may be it should the other way round:**
1. incoming PGN contents speed & angle
2. it get's parsed and it is written 
´´´
             update->setEnvironmentWindSpeedOverGround(windSpeed);
             update->setEnvironmentWindDirectionTrue(windAngle);
´´´
3. The NMEA0183 writing function has to wait for update on Angle and then it can take the new speed value for this sentence. 

If this logic is right, then we have to carefully watch, which value is the last coming in and this one triggers the following calculations.

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: I do not know if heading of more of 360° is allowed.
In AIS it would be treated as not valid.

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: Changed the code, but
´case  N2kWind_Apparent:´  is not working.
Timo has updated the library already and I made an update in Platformio too.
But I think due to platformio.ini it is called the old lib?

ronzeiller:
As discussed above:
validation of RudderAngle cancelled
if turned to switch + NORMALIZE_ANGLE implemented (Macro is in SKUnits.h at the moment)
´´´
if (windAngle >M_PI) windAngle = NORMALIZE_ANGLE(windAngle);
update->setEnvironmentWindAngleTrueGround(windAngle);
´´´
could also made directly like
´´´
update->setEnvironmentWindAngleTrueGround(NORMALIZE_ANGLE(windAngle));
´´´
I do not know which is better concerning CPU load

For other things (like generally validation question) I wait for your answer.


For the transducer thing:
https://github.com/sarfata/kbox-firmware/pull/84/commits/a1c89b51898fc2164e3df6743004475050fb34ae

Not clear why you dropped the 
´ else if (offset >**=** 0) { ´ ?

src/common/signalk/SKNMEAVisitor.cpp
sarfata: You are absolutely right! Sorry! Yes you do need the `!u.hasEnvironmentWindSpeedApparent()` validation.
My logic is that if the data comes through one PGN or one NMEA sentence, all the information should be transformed into one SignalK Delta message (one SKUpdate object). This is what the official nmea2000 tool and iKommunicate do.

For the angles, I think we should try to be as tolerant as possible and accept values that are not normalized. We should always send normalized values (0 to 2pi for directions, -pi to pi for angles).

src/common/signalk/SKNMEA2000Parser.cpp
sarfata: In platformio.ini you need to change the commit number we are referencing to:
```
  https://github.com/sarfata/NMEA2000#feature/fix-build-for-mac
```
becomes:
```
  https://github.com/ttlappalainen/NMEA2000#0aea81d 
 ```

It will conflict because I already did it on develop once before Timo merged. I will do it for you.

File: src/common/signalk/SKNMEAVisitor.cpp (hunk 27 above)
Comment (ronzeiller) [reply src/common/signalk/SKNMEAVisitor.cpp]: Good morning :-)

May be a function which normalizes angles and calculate +/- angle, something like:
```
void directionToAngle(double &angle) {
    NORMALIZE_ANGLE(angle);
    if (angle > M_PI ) angle += - M_2xPI;
}
```

**Comment by sarfata** [comment]
> As discussed above:
> - validation of RudderAngle cancelled
> - if-statement turned to switch 
> 
> - NORMALIZE_ANGLE should it not be?
> `  #include NORMALIZE_ANGLE(x)   fmod( x + M_PI * 2, M_PI * 2 )`

I think you are right. We need to verify what `fmod()` does for negative numbers.

> But where should angles > 360° coming from?

I think a "simple" nmea2000 device could be sending values that are not properly normalized. Maybe because they are a bit buggy or a bit dumb.

> And, if there are degrees coming instead of radians, then normalize would not help?

Yes you are right. If there are degrees instead of radians it's not going to work for sure. There is nothing we can do about this.

Thinking out  loud: Now that you mention degree, I realize that maybe it's not a good idea to try to accept values that are greater than 2 PI because if a buggy device sends value in degrees we will try to reduce them to a value in radian and it will be gibberish anyway. So maybe we should just reject anything that is not valid.

There is a surprising amount that has been written about this very same problem:
 - https://en.wikipedia.org/wiki/Robustness_principle
 - https://softwareengineering.stackexchange.com/questions/12401/be-liberal-in-what-you-accept-or-not
 - http://www.codingindelphi.com/blog/obey-postels-law-in-your-code/

To summarize:
 - Some people say it's better to accept anything (be robust, "postel's law")
 - Some people say it will make bugs harder to detect and will lead to weird behaviors

After reading this, I am changing my mind and I think we should follow your first proposal: be strict in what we accept. If we find a device on the market that is buggy then we will verify what it's doing and make sure we are compatible but trying to be compatible with bugs that we have not yet seen is probably a mistake.

I think this is a very important principle so I have started a new wiki page to keep track of that kind of decision: https://github.com/sarfata/kbox-firmware/wiki/Coding-Guidelines.

> 
> I did now for the moment:
> ```
> if (windAngle >M_PI) windAngle += -2* M_PI;
> update->setEnvironmentWindAngleTrueGround(windAngle);
> ```

I think you updated this with normalizeAngle which is great!

> in SKUnit.h:
> I do not know if it is better concerning CPU load to make
> `#define SKDegToRad(x) x * 0.017453292519943`
> instead of
> `#define SKDegToRad(x) x * 2 * M_PI / 360`

There is no CPU difference at all because it's a constant value and the compiler will calculate it for you when we are building KBox instead of doing the maths at runtime. So you can keep the more explicit version which is much easier to understand, and does not need a comment next to it.

> For other things (like generally validation question) I wait for your answer.
> 
> 
> For the transducer thing:
> https://github.com/sarfata/kbox-firmware/pull/84/commits/a1c89b51898fc2164e3df6743004475050fb34ae
> 
> Not clear why you dropped the 
> ` else if (offset >**=** 0) { ` ?

Sorry! I forgot to explain. This is so that this test would pass:
https://github.com/sarfata/kbox-firmware/blob/develop/src/test/signalk/SKNMEA2000ParserTest.cpp#L62

This test says that if we receive only the depth, without information about the position of the transducer we should not send the other signal k keys. I think this is the right thing to do. I know I told you before to do this but I had not thought about this case. Another good example of the usefulness of tests.

File: src/common/signalk/SKNMEA2000Parser.cpp (hunk 25 above)
Comment (sarfata) [reply src/common/signalk/SKNMEA2000Parser.cpp]: Ok.  As discussed let's keep it like this.

File: src/common/signalk/SKNMEA2000Parser.cpp [hunk 29]
```diff
@@ -72,34 +231,37 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
-    SKSource source = SKSource::sourceForNMEA2000(input, msg.PGN, msg.Priority, msg.Source);
-    update->setSource(source);
+      SKSource source = SKSource::sourceForNMEA2000(input, msg.PGN, msg.Priority, msg.Source);
+      update->setSource(source);
 
-    update->setEnvironmentDepthBelowTransducer(depthBelowTransducer);
+      update->setEnvironmentDepthBelowTransducer(depthBelowTransducer);
 
-    if (!N2kIsNA(offset)) {
+      if (N2kIsNA(offset)) {
```
sarfata: I think we do not need to set the offset to 0 here because we are testing to see if it's NA first. We can remove these 4 lines.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -109,13 +271,143 @@ const SKUpdate& SKNMEA2000Parser::parse128267(const SKSourceInput& input, const
...
+  double latitude;
+  double longitude;
+
+  if (ParseN2kPositionRapid(msg, latitude, longitude)) {
+    SKUpdateStatic<2> *update = new SKUpdateStatic<2>();
+    update->setTimestamp(timestamp);
+
+    SKSource source = SKSource::sourceForNMEA2000(input, msg.PGN, msg.Priority, msg.Source);
+    update->setSource(source);
+    update->setNavigationPosition(SKTypePosition(latitude, longitude, 0));
```
sarfata: Note for myself: I think we should use some form of NaN value here for altitude. I am not sure which one but this is something we will have to test later. SignalK tools do not send altitude when they do not get it.

src/test/signalk/SKNMEA2000ParserTest.cpp
```diff
@@ -83,5 +83,78 @@ TEST_CASE("SKNMEA2000Parser: Basic tests") {
...
+    CHECK( update.getEnvironmentWindSpeedApparent() == 12.4 );
+    CHECK( update.getEnvironmentWindAngleApparent() == Approx(SKDegToRad(-30)) );
+  }
+
+  SECTION("130306: Ground Wind Speed and Angle TWS, TWA port") {
+    SetN2kWindSpeed(msg, 0, 12.4, SKDegToRad(359), N2kWind_True_boat);
+    const SKUpdate &update = p.parse(SKSourceInputNMEA2000, msg, SKTime(0));
+    CHECK( update.getSize() == 2);
+    CHECK( update.getEnvironmentWindSpeedOverGround() == 12.4 );
+    CHECK( update.getEnvironmentWindAngleTrueGround() == Approx(SKDegToRad(-1)).epsilon(0.0001) );
```
sarfata: Added epsilon here because this test did not pass.

We lose some precision when we convert data to NMEA2000 (because NMEA2000 uses less bytes for double - and because the library automatically rounds values to 0.0001 for 2bytesdouble).

src/common/signalk/SKNMEAVisitor.cpp
sarfata: good morning :) 

As I said in my other big comment, I think you were right and we should be strict. Better detect problems early than later. Sorry for changing my mind :disappointed: 

Thanks for a great discussion that led me to change my mind! :1st_place_medal:

ronzeiller:
Yes, I like the tests more and more :thumbsup:

I put the Approx and epsilon thing to heading and rudder, as we handle angles there too.

A small thing. You code:

..```
..
  }
  else {
    DEBUG("Unable to parse N2kMsg with PGN %i", msg.PGN);
    return _invalidSku;
  }

```

And I did:
 ```
 }
  DEBUG("Unable to parse N2kMsg with PGN %i", msg.PGN);
  return _invalidSku;
  }
```
The outcome is the nearly same (I have always a return, even if something happens in the if-statement, but may be it is not necessary.)
It should probably the same style in the whole SKNMEA2000Parser.cpp
What is your opinion?

I made an update, hopefully with all discussed improvments

sarfata:
> A small thing. You code:
> 
> ```
> ...
>   }
>   else {
>     DEBUG("Unable to parse N2kMsg with PGN %i", msg.PGN);
>     return _invalidSku;
>   }
> ```
> 
> And I did:
>  ```
>  }
>   DEBUG("Unable to parse N2kMsg with PGN %i", msg.PGN);
>   return _invalidSku;
>   }
> ```
> The outcome is the nearly same (I have always a return, even if something happens in the if-statement, but may be it is not necessary.)
> It should probably the same style in the whole SKNMEA2000Parser.cpp
> What is your opinion?

I like having no else. I will remove it in the other functions.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -72,34 +231,33 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
 
-    SKSource source = SKSource::sourceForNMEA2000(input, msg.PGN, msg.Priority, msg.Source);
-    update->setSource(source);
+      SKSource source = SKSource::sourceForNMEA2000(input, msg.PGN, msg.Priority, msg.Source);
+      update->setSource(source);
 
-    update->setEnvironmentDepthBelowTransducer(depthBelowTransducer);
+      update->setEnvironmentDepthBelowTransducer(depthBelowTransducer);
 
-    if (!N2kIsNA(offset)) {
```
sarfata: I was wrong when I told you we did not need this anymore. Removing this broke the tests. I am re-adding it.

sarfata:
@ronzeiller Please take a look one last time. If you are happy with my changes and `SKNormalizeAngle()`, `SKNormalizeDirection()` then I think we can merge!

ronzeiller:
> @ronzeiller Please take a look one last time. If you are happy with my changes and SKNormalizeAngle(), SKNormalizeDirection() then I think we can merge!

Great to have them both! 

At `SKNormalizeDirection()`
you have:`[-PI,2*M_PI)`
I think it should be:  `* Normalizes any angle in radians to the range [-M_PI, M_PI)` 

I did not know, that M_PI is already "wind from port"?

sarfata:
> At `SKNormalizeDirection()`
> you have:`[-PI,2*M_PI)`
> I think it should be:  `* Normalizes any angle in radians to the range [-M_PI, M_PI)` 

Good catch! I have fixed this now. I looked online and this seems to be the most common convention. This is why I have used it.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -71,9 +73,77 @@ void SKNMEAVisitor::visitSKNavigationAttitude(const SKUpdate &u, const SKPath &p
...
+  float windSpeed;
+  windAngle = v.getNumberValue();
+
+  // Look if we have an update for speed too
+  if (!u.hasEnvironmentWindSpeedApparent()) return;
+
+  windSpeed = u.getEnvironmentWindSpeedTrue();  // in m/s
+
+  NMEASentenceBuilder sb( "II", "MWV", 5);
+  sb.setField(1, SKRadToDeg(windAngle), 2 );
```
sarfata: I just realized that when we parse the angle from NMEA2000 we do not apply `SKNormalizeDirection()` so it's possible that the angle is negative (-35 degrees for example). However `MWV` specifically calls for an angle between 0 and 360 so I think we should normalize to a direction here. Do you agree?

ronzeiller: Yes, true! Added and pushed.

In the HDM (Heading Magnetic) should there be a NormalizeDirection too?

And in the NMEA2000Parser itself I miss your nice macro?
```
case N2kWind_True_North:
  // Ground Wind Direction
  update->setEnvironmentWindDirectionTrue(windAngle); --> SKNormalizeDirection()

case N2kWind_Magnetic:
  // Ground Wind Direction referred to magnetic north
  update->setEnvironmentWindDirectionMagnetic(windAngle); --> SKNormalizeDirection()
    
```

sarfata:
Oh @ronzeiller I am so sorry, I did not see your last message! Taking a look now.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -71,9 +73,76 @@ void SKNMEAVisitor::visitSKNavigationAttitude(const SKUpdate &u, const SKPath &p
...
+void SKNMEAVisitor::visitSKEnvironmentWindAngleApparent(const SKUpdate &u, const SKPath &p, const SKValue &v) {
+
+  float windAngle;
+  float windSpeed;
+  windAngle = v.getNumberValue();
+
+  // Look if we have an update for speed too
+  if (!u.hasEnvironmentWindSpeedApparent()) return;
+
+  windSpeed = u.getEnvironmentWindSpeedTrue();  // in m/s
```
sarfata: That is a nasty bug that I found because I was adding some more tests. Should be `WindSpeedApparent` here.

src/common/signalk/SKNMEAVisitor.cpp
```diff
@@ -71,9 +73,76 @@ void SKNMEAVisitor::visitSKNavigationAttitude(const SKUpdate &u, const SKPath &p
...
+//             VW Velocity Sensor, Mechanical?
+//  also seen: IIVWR
+//
+//  NMEA0183  MWV Wind Speed and Angle  --> recommended
+//              1  2  3  4
+//              |  |  |  |
+//      $--MWV,x.x,a,x.x,a*hh
+//  1) Wind Angle, 0 to 360 degrees
+//  2) Reference, R = Relative, T = True
+//  3) Wind Speed Units: K = km/s, M = m/s, N = Knots
```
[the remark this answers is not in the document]
ronzeiller: Sorry, some kind of blackout....

ronzeiller:
After all, it seems that it would have been easier, if just you have made all :-)))
But, I learned so much during this work, that future PRs should be easier for you ;-)

⛵️

sarfata:
No it would not have been easier if I had done it alone ;) And I think this stuff was definitely way more complicated than one would think ... 

It was great to have someone else to collaborate with. Thanks a lot!

PR #88, opened 2017-12-08
Feature/support wind to n2k

sarfata:
Adds conversion from SignalK to NMEA2000 and a bunch of tests to go with it. I have tested that apparent wind is properly converted and displayed.

I had a few NMEA sentences with True wind that are converted but the displays seem to ignore them anyway.

@ronzeiller do you want to take a look at this PR before I merge it?
(it looks big now because I made it on top of yours. Once we merge your branch, there is only one commit here)

### DIFF
--- a/src/common/signalk/SKNMEA2000Visitor.cpp
+++ b/src/common/signalk/SKNMEA2000Visitor.cpp
@@ -22,11 +22,13 @@
   THE SOFTWARE.
 */
 
+#include "SKNMEA2000Visitor.h"
+
 #include <N2kMsg.h>
 #include <N2kMessages.h>
-#include "SKNMEA2000Visitor.h"
 #include "SKUpdate.h"
 #include "SKValue.h"
+#include "SKUnits.h"
 
 SKNMEA2000Visitor::SKNMEA2000Visitor() {
 }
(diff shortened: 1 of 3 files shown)


sarfata:
Also, I verified that NMEA2000 messages are sent properly. I think you said that sometimes NMEA2000Visitor is not called. Where was that? It does work here for sure.

ronzeiller:
@Sarfata
The `visitSKEnvironmentWind....` functions look all right.

For the 
`generateWind(double windAngle, double windSpeed, tN2kWindReference windReference) {....`

I have a question:
you write 
```
// We need to normalize wind to a direction because it is stored as un
// unsigned double.
```
Didn't we store signed and normalized directions?

Are you sure you need `SKNormalizeDirection(windAngle)` for all `windAngle`?
As there are coming angles and directions mixed.

....but why the tests get passed successfully then? :confused:

ronzeiller:
> Also, I verified that NMEA2000 messages are sent properly. I think you said that sometimes NMEA2000Visitor is not called. Where was that? It does work here for sure.

Yes, I mentioned this some time ago. 
There was nothing triggering when a PGN was coming in from the bus.
But with the new `visitSKEnvironmentWind....-`functions it should work. 

So the next steps are to bring the nicely built NMEA0183 sentences somehow to the Serial output.

You told me some month ago (for KBox master branch) to do a simple Serial.print 
but this would not cover the different NMEA_n outputs, the different baud rates on those outputs, and,........

Any plans for this?

sarfata:
> For the 
> `generateWind(double windAngle, double windSpeed, tN2kWindReference windReference) {....`
> 
> I have a question:
> you write 
> ```
> // We need to normalize wind to a direction because it is stored as un
> // unsigned double.
> ```
> Didn't we store signed and normalized directions?

We store signed internally in SignalK format but NMEA2000 stores in unsigned format.

When we parse NMEA2000 messages we need to make sure we convert 330 to -30 - and then when we generate NMEA2000 we need to make sure we convert -30 to 330.

We have tests for both scenario:
- https://github.com/sarfata/kbox-firmware/pull/88/files#diff-c43b3193341d41d532c69e5c40c51c66R169
- https://github.com/sarfata/kbox-firmware/blob/develop/src/test/signalk/SKNMEA2000ParserTest.cpp#L141 (you wrote this one ;)

> 
> Are you sure you need `SKNormalizeDirection(windAngle)` for all `windAngle`?
> As there are coming angles and directions mixed.
> 
> ....but why the tests get passed successfully then? :confused:

Yes we need it because angles and directions are always represented as one number between 0 and 360 degrees by NMEA2000 (which is something I learnt yesterday when I realized they store in an unsigned double).

Thanks for taking a look. I will merge this one too now.

PR #89, opened 2017-12-09
IMUMonitorPage added

ronzeiller:
Hi Thomas,
started a new IMUMonitorPage PR as draft for discussion:

1. introduced a config file. The idea is to start config with something which will grow, also to get a collection and an overview. When there is a solution found later, ini-file, JSON or whatsoever it will be easy to change the setting variables already in use.
2. I do not know how to combine/extend IMUService and IMUMonitorPage to set/get cal variables.
friend class? Could you help here please?
3. The plan is to build in a filter for the display as discussed earlier into the Page. Is working already in an experimental version here.
4. For different mounting positions of KBox: I would like to start in the IMUService and not in the driver (as discussed earlier) because beside of axis and sign turning it will also be necessary to invert pitch or heel or transform angles, as done already in the Service.
I would like to start with our mounting positions (if somebody needs another, no problem). (Is already working here, I just wanted to pull something to discuss before the PR is too large....)
5. Last but not least I plan to write the calibration into the EEPROM and load it at start.
And I would like to introduce an offset into pitch and heel for zero-setting by a long button-press. Reading, writing and offset is working already. As soon I can get/set variables from the Page -> Service I can implement it. 
For storing the cal-datas: I thought of first storing, if all calibration is 3 and then check every 10 minutes? What is your opinion?

The page is working already (without cal display and without filter)


src/host/services/IMUService.cpp
```diff
@@ -55,9 +56,13 @@ void IMUService::loop() {
...
   heading = fmod(eulerAngles.x() + 270, 360);
 
-  if (sysCalib == 3) {
-    update.setNavigationAttitude(SKTypeAttitude(/* roll */ SKDegToRad(roll), /* pitch */ SKDegToRad(pitch), /* yaw */ 0));
+  if (magCalib >= cfHdgMinCal) {
     update.setNavigationHeadingMagnetic(SKDegToRad(heading));
     _skHub.publish(update);
+  }
+	
+	if (accelCalib >= cfHeelPitchMinCal && gyroCalib >= cfHeelPitchMinCal) {
```
sarfata: I think there is a indentation issue here. Can you make sure your editor is configured to replace tabs with spaces?

src/host/main.cpp
```diff
@@ -102,6 +104,11 @@ void setup() {
 
   BatteryMonitorPage *batPage = new BatteryMonitorPage(skHub);
   mfd.addPage(batPage);
+	
+	#ifdef IMU_MONITOR_PAGE
+    IMUMonitorPage *imuPage = new IMUMonitorPage(skHub);
```
sarfata: To allow `IMUMonitorPage` to have more access to `IMUService`, pass a reference to the `IMUService` when you create `IMUMonitorPage`: 

```
IMUMonitorPage *imuPage = new IMUMonitorPage(skHub, *imuService);
```

And then in `IMUService.h` you can declare new functions for example:
```
public:
  void getLastValues(int &accelCalibration, double &pitch, double &heel, int &magCalibration, double &heading);
  void saveAccelCalibration();
  void resetAccelCalibration();
  void saveMagCalibration();
  void resetMagCalibration();
```
(and anything else you need)

This way, the `imuPage` will have more access to imuService but we will keep the imu logic inside the imu service.

lib/KBoxHardware/src/KBoxConfig.cpp
```diff
@@ -0,0 +1,52 @@
+/*
```
sarfata: I don't think we should put this in KBoxHardware because this folder is only for hardware support. The config is very specific to what kbox does so I would just put this into `src/common/config` (so it can be shared with the wifi module too).

lib/KBoxHardware/src/KBoxConfig.cpp
```diff
@@ -0,0 +1,52 @@
...
+  THE SOFTWARE.
+*/
+
+#include "KBoxConfig.h"
+
+// ---------------------------------------------------------------------------
+//           Use of internal Sensors
+// ---------------------------------------------------------------------------
+// [sensors]
+bool cfUseHdgFromBusIfExists			= true;   // Get Heading from NMEA2000?
```
sarfata: I think this should be a class with methods. To begin, we can just hardcode the values in the methods inside the `.h` file but in the future we can re-implement this class and read from JSON.

sarfata:
This implements #10

lib/KBoxHardware/src/KBoxConfig.cpp
ronzeiller: ok, done.
Just to keep in mind, that (as far as I know) if the config is in the src part then libraries can not access it, which means that we will *have to* do the axis- and sign thing in IMUService and not in the driver.
(I think for now it is perfect, and it could easily be changed later, if needed)

lib/KBoxHardware/src/KBoxConfig.cpp
ronzeiller: You mean to make them as private variables and implement a getXyZ() for all of them?

Could you propose the name of the class to stay within your naming concept?

src/host/main.cpp
ronzeiller: This is looks nice, but I do not know how to declare 
```IMUMonitorPage(SKHub& hub, IMUService* ims);```
this gives me an error "redefinition of class IMUService"
how to do this right?

File: src/host/services/IMUService.cpp (hunk 1 above)
Comment (ronzeiller) [reply src/host/services/IMUService.cpp]: Thanks, I was trying Netbeans as IDE because of it's internal versioning which I find great. Will have to take more care in configuring it.

**Comment by ronzeiller** [comment]
Got rid of the redefinition error , but now I have the ```imuService.getLastValues```in the constructor of IMUMonitorPage and not in the 
```void IMUMonitorPage::updateReceived(const SKUpdate& up)```

Could you please have a look at this?

Now the concept is: the values for the monitorPage are triggered by ```updateReceived(const SKUpdate& up)```
which means the updates are coming by the interval of the imuService.
In the IMUPage would be the filter and a elapsed time thing for bringing down the values to display-frequeny.

Just thinking of: would it be better to make the **display of data** (at least the Voltage, Baro, IMU, may be wind and other pages) an **interval task?**
This would mean, that the filter would be in the IMUService (which would be there anyway if you might switch to a better IMU sensor for future KBox Hardware releases e.g. this one: https://www.adafruit.com/product/3463 which has no integrated fusion as the BNO055)
The other advantage of filtering in high frequency is to get out noise without visual time delay. (which we certainly have at low frequencies, you can see it quite a lot in openCPN!)

The Page loop would then just call an actual value for the display (completely  independent from the Service-frequency.) Same for different other (low speed) outputs....

**The other things:**
1. config settings will change as discussed
2. two Service defines added. 
The aim is, that users could define their services needed for a special hardware. 
E.g. it would then be possible to comment out several services to make a logging device with a T3.2 or 3.6 + CAN without display, wifi, baro, imu
Or just to switch off Baro and IMU if not needed and so on....
Or with a special (less equipped and lower priced) KBox as N2k->NMEA0183 Serial/WiFi-converter.

Would you like this?

**Comment by sarfata** [comment]
> Just thinking of: would it be better to make the display of data (at least the Voltage, Baro, IMU, may be wind and other pages) an interval task?
> This would mean, that the filter would be in the IMUService (which would be there anyway if you might switch to a better IMU sensor for future KBox Hardware releases e.g. this one: https://www.adafruit.com/product/3463 which has no integrated fusion as the BNO055)
> The other advantage of filtering in high frequency is to get out noise without visual time delay. (which we certainly have at low frequencies, you can see it quite a lot in openCPN!)
> 
> The Page loop would then just call an actual value for the display (completely independent from the Service-frequency.) Same for different other (low speed) outputs....

I think I agree but let me restate this to make sure we are on the same page:

* `IMUService` should process data at high speed (ideally, as soon as it's available) and will apply the filtering and send the data to the `SKHub`
* Rewriting `IMUPage` as a interval based task makes sense because you know data will always be available from our IMU **and** you need calibration data that is not included in `SKUpdate`.
  * The only concern I have with this is that IMUPage will always only show value from KBox internal sensor, and not from another device on the network but that is completely fine in this case.
  * The MFD task already takes care of redrawing the screen. You do not need (and should not) create a new interval service. Instead, override the `processEvent(const TickEvent &e)` method (from `Page.h`) to update the text fields every x ms.



Regarding config, let me put something together too so we can compare our approaches and it will be more explicit for me to write code than explain this in Github comments ;)

I want to arrive at a configuration system based on a file (INI or JSON) that reads configuration when KBox starts. I do not want users (or me ;) ) to have to re-compile just to change some basic options.

I understand why you have the `#define` but it forces us into a "build-time" setting. For pages and services this might be the right option. For other options, it will not be the right option. I think we should try to make things configurable at runtime first, and revert to build time if we are forced to for some reason.

For example, in the case of a KBox variant, I think the built time configuration is the only thing that makes sense but we can do it with a dedicated target and a build time `-DKBOX_MINI_VARIANT`.

File: src/host/pages/IMUMonitorPage.h [hunk 5]
```diff
@@ -0,0 +1,51 @@
...
+
+#include "common/ui/Page.h"
+#include "common/ui/TextLayer.h"
+#include "common/signalk/SKHub.h"
+#include "common/signalk/SKSubscriber.h"
+#include "services/IMUService.h"
+
+class IMUMonitorPage: public Page, public SKSubscriber {
+  private:
+    TextLayer *_hdgTL, *_heelTL, *_pitchTL, *_calTL;
```
sarfata: Declare a new private reference here to keep the reference to the `IMUService`:

```
    IMUService &_imuService;
```

src/host/pages/IMUMonitorPage.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+#include <stdio.h>
+#include "common/signalk/SKUpdate.h"
+#include "IMUMonitorPage.h"
+#include "common/config/KBoxConfig.h"
+#include <KBoxLogging.h>
+#include "signalk/SKUnits.h"
+
+
+IMUMonitorPage::IMUMonitorPage(SKHub& hub, IMUService &imuService) {
```
sarfata: You need to keep that reference to `imuService` so you can use it later.

Just change this line to:
```
IMUMonitorPage::IMUMonitorPage(SKHub& hub, IMUService &imuService) : _imuService(imuService) {
```

This means that when the class is initialized, it will store the reference in the private property `_imuService` (which we need to declare below).

src/host/pages/IMUMonitorPage.cpp
```diff
@@ -0,0 +1,110 @@
...
+  _pitchTL = new TextLayer(Point(col2, row4), Size(20, 20), "--", ColorWhite, ColorBlack, FontLarge);
+
+  addLayer(_hdgTL);
+  addLayer(_calTL);
+  addLayer(_heelTL);
+  addLayer(_pitchTL);
+
+  hub.subscribe(this);
+
+  imuService.getLastValues(_accelCalibration, _pitch, _heel, _magCalibration, _heading);
```
sarfata: We are saving it now so you can refresh this whenever you want. For example, in the method `processEvent(const TickEvent &e);`.

sarfata: One of the advantage of using `TickEvent` is that this page will only display when it's on the screen. Not when it's hidden.

ronzeiller:
> I think I agree but let me restate this to make sure we are on the same page:
> 
> IMUService should process data at high speed (ideally, as soon as it's available) and will apply the filtering and send the data to the SKHub
> 

Yes

- The idea behind was, that any sensor software should deliver fine datas.
- We have to make a difference of high speed data filtering for a specific sensor, which is in the sensorService.
- And we have different outputs. 
Where a display should make it's own (selectable) display filter/damping. 
This is how it is made on every MFD on the boat. 1.)


The question for later on will be: how and where it will be done for a slow serial output......?

> Rewriting IMUPage as a interval based task makes sense because you know data will always be available from our IMU and you need calibration data that is not included in SKUpdate.

If I understood it right, the display is based on the eventTicks, which is already an interval by `processEvent(const TickEvent &e) `

> The only concern I have with this is that IMUPage will always only show value from KBox internal sensor, and not from another device on the network but that is completely fine in this case.

I think we could make this dependent of a config-setting 
`"take internal sensor for HDG"` and `"take internal sensor for heel/pitch"` 2.)


> Regarding config, let me put something together too so we can compare our approaches and it will be more explicit for me to write code than explain this in Github comments ;)
> 
> I want to arrive at a configuration system based on a file (INI or JSON) that reads configuration when KBox starts. I do not want users (or me ;) ) to have to re-compile just to change some basic options.

Yes, this was my intention too.
This sounds great, I will love to wait for your approaches!!


> I understand why you have the #define but it forces us into a "build-time" setting. For pages and services this might be the right option. For other options, it will not be the right option. I think we should try to make things configurable at runtime first, and revert to build time if we are forced to for some reason.

> For example, in the case of a KBox variant, I think the built time configuration is the only thing that makes sense but we can do it with a dedicated target and a build time -DKBOX_MINI_VARIANT.

100% d'accord
Yes, this was the idea, just for services and pages and nothing which could be made at runtime.
This #define was just my minimalistic approach with my low level skills of C++ 
(knowing not enough to do this target thing... ;-) )

Foot notes:
1.) So it all depends of: what happens if we take and display a value, let's say every second.
Does it jump around and become unreadable?
E.g. May be BNO055's fusion output is that good, that it doesn't need any display-filter at all?

2.) If the setting is `"take heading/heel&pitch from N2k-Bus, an own external IMU-sensor on serial input" ` then the page should show that values.

src/host/pages/IMUMonitorPage.h
ronzeiller: Great! Worked fine!
Thank You,
update will come soon :-)

ronzeiller:
Left to do:
1. change config setting variables to new system
2. Check if Cal is maximum (=3) and save calibration to EEPROM from time to time
3. invent some display smoothing according to IMUService interval (later to be set by config setting)
4. Do some more mounting position variations ( mine :smile: )
5. waiting for your highly appreciated comments :relaxed:

CaptainRon47:
Is any of this in place in the 'new' Master (used to be 'develop') firmware?  I have been hoping to have the display pages scroll based on a time sequence and to have new pages (barometer, IMU, wind data, etc) added.  This would free up the current button for another function.  My thought being that there would be separate 'constants' set of data the could be uploaded based on the button push.  These constants would be specific for your boat.  I think this is what you are talking about, but I don't know how far along you are on any of this.
thanks,
ron

ronzeiller:
Hi @CaptainRon47 
> Is any of this in place in the 'new' Master (used to be 'develop') firmware?
> I have been hoping to have the display pages scroll based on a time sequence and to have new pages (barometer, IMU, wind data, etc) added. This would free up the current button for another function. 

If the code is approved it will come an IMU-Page with Heading, heel, pitch and cal as quality display.

An own page for pressure would be nice in deed. With a pressure-graph of some time period....
My BMP280 is not working, so I do not have pressure here :-(

But may be I start with wind a page and a graph of lasting true wind (strength and angle)

I think that time based scrolling is not the solution I would prefer.
But may be it is enough to do scrolling by 'turning the button'. 

Then we could have 'click', 'long click' and may be 'double click' for actions

> My thought being that there would be separate 'constants' set of data the could be uploaded based on the button push. These constants would be specific for your boat. I think this is what you are talking about, but I don't know how far along you are on any of this.

There is planned to do a sort of config setup which can be stored on SD-Card.
What kind of 'constants' you are thinking of?

Ronnie

CaptainRon47:
Ron,
Great, I never realized that rotating the button was an option!  That would be fine, the only reason I would like the instruments at the KBox display is to make sure everything is working.  I currently mount it where it cannot be normally seen.  The constants are just the specifics for each boat; how you mounted the IMU (which is pitch and roll), how you want units displayed (m, ft, barometric pressure, etc), the Wfi settings, the NMEA speeds.  This is not a big deal as once you have them set you don't really change them.  So even just having them at one place in the code would be fine.

thanks,
RonS

src/common/ui/Event.h
```diff
@@ -40,8 +40,16 @@ class Event {
 };
 
 enum ButtonEventType {
+  // Those two events are always sent when button goes up/down
   ButtonEventTypePressed,
-  ButtonEventTypeReleased
+  ButtonEventTypeReleased,
+
+  // Those events are also sent when a single click, or a double click happen
+  ButtonEventTypeClick,
```
sarfata: Very very nice :+1:

src/common/config/KBoxConfig.h
```diff
@@ -0,0 +1,67 @@
+/*
```
sarfata: I have a separate PR coming up for configuration. I worked on it yesterday but did not have time to finish today. Maybe we can remove config stuff from this PR and just focus on the IMU with one fixed position?

src/host/pages/IMUMonitorPage.cpp
```diff
@@ -0,0 +1,107 @@
...
+  _rollTL  = new TextLayer(Point(col1, row4), Size(20, 20), "--", ColorWhite, ColorBlack, FontLarge);
+  _pitchTL = new TextLayer(Point(col2, row4), Size(20, 20), "--", ColorWhite, ColorBlack, FontLarge);
+
+  addLayer(_hdgTL);
+  addLayer(_calTL);
+  addLayer(_rollTL);
+  addLayer(_pitchTL);
+
+  hub.subscribe(this);
+
```
sarfata: nitpick: we should avoid extra blank lines at the beginning or end of functions. You have a few in this file.

src/host/pages/IMUMonitorPage.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include "common/ui/Page.h"
+#include "common/ui/TextLayer.h"
+#include "common/signalk/SKHub.h"
+#include "common/signalk/SKSubscriber.h"
+#include "services/IMUService.h"
+
+class IMUMonitorPage : public Page, public SKSubscriber {
+	private:
+		TextLayer *_hdgTL, *_rollTL, *_pitchTL, *_calTL;
```
sarfata: The indentation is wrong here. Should 2 space for private and then 4 for the properties and methods.

src/host/services/IMUService.cpp
```diff
@@ -23,41 +23,138 @@
...
 
-  /* if orientation == MOUNTED_ON_PORT_HULL */
-  double roll, pitch, heading;
-  roll = eulerAngles.z();
-  pitch = eulerAngles.y();
-  heading = fmod(eulerAngles.x() + 270, 360);
+  /* if orientation == MOUNTED_ON_STARBOARD_HULL */
+  _roll = eulerAngles.z() + _offsetRoll;
+  _pitch = eulerAngles.y() + _offsetPitch;
+  _heading = fmod(eulerAngles.x() + 270, 360);
```
sarfata: We should always store data internally in radians. Otherwise I think we risk confusing ourselves...

src/host/services/IMUService.cpp
```diff
@@ -23,41 +23,138 @@
...
+
+  if (_magCalib >= cfHdgMinCal) {
+    update.setNavigationHeadingMagnetic(SKDegToRad(_heading));
+    _skHub.publish(update);
+  }
 
-  if (sysCalib == 3) {
-    update.setNavigationAttitude(SKTypeAttitude(/* roll */ SKDegToRad(roll), /* pitch */ SKDegToRad(pitch), /* yaw */ 0));
-    update.setNavigationHeadingMagnetic(SKDegToRad(heading));
+	if (_accelCalib >= cfHeelPitchMinCal && _gyroCalib >= cfHeelPitchMinCal) {
```
sarfata: indent problem here.

src/host/services/IMUService.cpp
```diff
@@ -23,41 +23,138 @@
...
+   sensor_t sensor;
+   adafruit_bno055_offsets_t newCalib;
+   bno055.getSensorOffsets(newCalib);
+
+   DEBUG("\n\nStoring calibration data to EEPROM...");
+
+   bno055.getSensor(&sensor);
+   bnoID = sensor.sensor_id;
+
+   EEPROM.put(_eeAddress, bnoID);
```
sarfata: Instead of writing directly to the EEPROM could you try to re-use https://github.com/sarfata/kbox-firmware/blob/master/src/host/util/PersistentStorage.cpp?

This way we have only one place where we manage what is stored in EEPROM and where. It will help us make sure that we are not overwriting our own data and if we want to choose to another type of storage (on another micro-controller that does not have EEPROM for example) then we can do it in one place only.

src/host/services/IMUService.h
```diff
@@ -21,20 +21,29 @@
   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
   THE SOFTWARE.
 */
-
+#pragma once
 #include <Adafruit_BNO055.h>
 #include "common/os/Task.h"
 #include "common/signalk/SKHub.h"
+#include <EEPROM.h>
```
sarfata: You only need this in the `.cpp` file so you should put it only there. Having unused `#include` just slows down compile times.

src/host/services/IMUService.h
```diff
@@ -21,20 +21,29 @@
...
 
 class IMUService : public Task {
   private:
     SKHub &_skHub;
     Adafruit_BNO055 bno055;
-    uint8_t sysCalib, gyroCalib, accelCalib, magCalib;
+    uint8_t _sysCalib, _gyroCalib, _accelCalib, _magCalib;
```
sarfata: :+1: thanks for fixing my code ;)

src/host/services/MFD.cpp
```diff
@@ -45,9 +45,26 @@ void MFD::processInputs() {
   if (button.update()) {
     if (button.fallingEdge()) {
       events.add(new ButtonEvent(ButtonEventTypePressed));
+      lastButtonDown = millis();
+      lastMaintainedEvent = millis();
     }
     else {
       events.add(new ButtonEvent(ButtonEventTypeReleased));
+
+      if (millis() - lastButtonDown > longClickDuration) {
```
sarfata: Really like this. Thanks a lot for adding this in.

lib/Adafruit_BNO055/Adafruit_BNO055.h
```diff
@@ -307,10 +307,13 @@ class Adafruit_BNO055 : public Adafruit_Sensor
     void  setSensorOffsets(const adafruit_bno055_offsets_t &offsets_type);
     bool  isFullyCalibrated(void);
 
+    // make public to write in imuService
+    bool  write8  ( adafruit_bno055_reg_t, byte value );
```
sarfata: Maybe a better way would be to add a method `setAxisRemapConfig()`. This would respect the principles of encapsulation better.

lib/Adafruit_BNO055/Adafruit_BNO055.cpp
```diff
@@ -113,12 +113,12 @@ bool Adafruit_BNO055::begin(adafruit_bno055_opmode_t mode)
 
   /* Configure axis mapping (see section 3.4) */
   //write8(BNO055_AXIS_MAP_CONFIG_ADDR, REMAP_CONFIG_P2); // P0-P7, Default is P1
-  write8(BNO055_AXIS_MAP_CONFIG_ADDR, 0b00001001); // P0-P7, Default is P1
```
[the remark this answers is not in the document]
ronzeiller: Hi,
for the BNO055 library, the original code is (line 33 ff):
```
/* Set the output units */
  /*
  uint8_t unitsel = (0 << 7) | // Orientation = Android
                    (0 << 4) | // Temperature = Celsius
                    (0 << 2) | // Euler = Degrees
                    (1 << 1) | // Gyro = Rads
                    (0 << 0);  // Accelerometer = m/s^2
  write8(BNO055_UNIT_SEL_ADDR, unitsel);
  */

  /* Configure axis mapping (see section 3.4) */
  /*
  write8(BNO055_AXIS_MAP_CONFIG_ADDR, REMAP_CONFIG_P2); // P0-P7, Default is P1
  delay(10);
  write8(BNO055_AXIS_MAP_SIGN_ADDR, REMAP_SIGN_P2); // P0-P7, Default is P1
  delay(10);
  */
````
The rest was by done by you.

Then you did some corrections to wire at the beginning of the Adafruit_BNO055.cpp
So I am afraid we are not back at the original yet.

File: lib/Adafruit_BNO055/Adafruit_BNO055.h (hunk 18 above)
Comment (ronzeiller) [reply lib/Adafruit_BNO055/Adafruit_BNO055.h]: May be we should write this to Adafruit, to let them make it, so we can use the original library?

File: src/common/config/KBoxConfig.h (hunk 9 above)
Comment (ronzeiller) [reply src/common/config/KBoxConfig.h]: great!!
Since I got some home works to do at my PR anyway :-)
But yes, we can remove all the config stuff, no problem.

File: src/host/services/IMUService.cpp (hunk 12 above)
Comment (ronzeiller) [reply src/host/services/IMUService.cpp]: You are definitely right, may be we should name the variable according to units.
I don´t know if it makes sense to calculate twice if just degrees are needed for something?

File: src/host/services/IMUService.cpp (hunk 13 above)
Comment (ronzeiller) [reply src/host/services/IMUService.cpp]: Will take care of all ident things, caused to open the code in Netbeans before configuring :-(

File: src/host/services/IMUService.cpp (hunk 14 above)
Comment (ronzeiller) [reply src/host/services/IMUService.cpp]: OK, makes a lot of sense. Will try to do it.

File: src/host/services/IMUService.h (hunk 15 above)
Comment (ronzeiller) [reply src/host/services/IMUService.h]: slowly, slowly I will learn this.... :-)

**Comment by CaptainRon47** [comment]
Is there a version of this I can download and try or do I have to wait until you merge it into the Master?  I tried to follow along (after the fact), but I am still NOT a programmer!  While you were making changes to Event.h why did you not add in an event for button rotation for future use?

File: src/host/services/IMUService.cpp (hunk 12 above)
Comment (sarfata) [reply src/host/services/IMUService.cpp]: I think consistency is important. I would rather convert twice (it's very basic) rather than start mixing up units.

File: lib/Adafruit_BNO055/Adafruit_BNO055.h (hunk 18 above)
Comment (sarfata) [reply lib/Adafruit_BNO055/Adafruit_BNO055.h]: That would be the right thing to do indeed :) Maybe in another PR ;) Let's tackle one problem at a time.

**Comment by sarfata** [comment]
@CaptainRon47 GitHub shows you at the top of this page that the origin of this PR is in `ronzeiller` account and the name of the branch is `rz_imu-page`.

image

So if you want to grab it, you need to go there:
https://github.com/ronzeiller/kbox-firmware/tree/rz_imu-page and click "Download". Thanks for learning GitHub with us!

File: src/host/pages/IMUMonitorPage.h (hunk 11 above)
Comment (ronzeiller) [reply src/host/pages/IMUMonitorPage.h]: repaired

File: src/host/services/IMUService.h (hunk 15 above)
Comment (ronzeiller) [reply src/host/services/IMUService.h]: repaired

File: src/host/pages/IMUMonitorPage.cpp (hunk 10 above)
Comment (ronzeiller) [reply src/host/pages/IMUMonitorPage.cpp]: fixed

**Comment by ronzeiller** [comment]
The Adafruit BNO055 library has to be (and was) changed for KBox anyway, so I extended the begin() function in the library *) and the IMUService for different mounting positions.

Other changes (hopefully) made as discussed.

*) may be we should fork the Adafruit BNO055 library, make all changes for KBox in a clear way and link to "our own" library?

**Comment by CaptainRon47** [comment]
FYI
I downloaded the rz firmware, updated PlatformIO, did a PI build received the following errors:
C/C++ project index rebuild failed - Could not load broken JSON: ...\.piolibdeps\_tmp_installing-1b5xs_-package\library.json  ?
host, program-esp, esp, mfg all built successfully but test and sktool failed due to something in NMEA2000-library (actisensreader.o and  n2kdevicelist.o )

UPLOADED HOST AND SEE NEW IMU PAGE WITH HEADING, CALIB, HEEL AND PITCH VALUES
THANK YOU!

**Comment by CaptainRon47** [comment]
The serial output (USB) has a stats, wifi connection status and Barometer output (no IMU or voltages)
the wifi output has voltages, heading and IMU NMEA0183 sentences but no Barometer

**Comment by ronzeiller** [comment]
@CaptainRon47 
May be you have old NMEA2000 libraries in your .pio environment and a "clean" would help to build all without errors.
@Sarfata is working on a config system of KBox. With this it will be soon easier to config all sorts of things, e.g. switching output of datas to USB (now it is debug only), to enable which NMEA0183 on  which interface, and......

Btw, how do you mount your KBox on your boat? (stb, port, hull,.....)

As soon as the IMUPage is merged into the KBox firmware master, I would suggest to switch back to this firmware.

FYI: A long button press on the IMUPage let you make zero heel/pitch.

**Comment by CaptainRon47** [comment]
Ronnie,
Thanks, I had not tried the long press and I will perform the 'clean' operation next time.
My Kbox is mounted inside the instrument panel (where it is safe and dry) with the display facing forward and the SD card on the top.  I switched the original roll and pitch and reversed their sign, the mounting did not require an offset.  I turned off the magnetic heading because I get that from my on board B&G Compass.  As I said, my changes were adhoc and done in KMessageNMEAVisitor.cpp right were the sentences were being built.

**Comment by CaptainRon47** [comment]
Ronnie,
Sorry I did forget to say that my instrument panel is on the stbd side of the boat.  So my kbox is on a stb bulkhead, with display facing forward and the SD card slot on the top (button on the bottom).

**Comment by ronzeiller** [comment]
@sarfata 
Could you please check this PR with hopefully all terms implemented as discussed.

- IMUMonitorPage if IMUService is enabled
- config for calibration values of heel/pitch and hdg. If calibration values of sensor >= config, values for hdg and/or heel/pitch will be sent to SKHub
- Values will always be shown at IMUMonitorPage (if Service is enabled), but if calibration is lower, then color is changing to red
- Button pressed long calculates an offset and switches heel/pitch to zero

IMUService extended:
- store calibration to BNO055 EEPROM function available but not used yet
- read calibration from BNO055 EEPROM at startup
- Adafruit library extended to configure axis and sign for different mounting positions. 

Later the code could be extended to show hdg and heel/pitch from boat sensors.

File: lib/Adafruit_BNO055/Adafruit_BNO055.cpp [hunk 20]
```diff
@@ -53,16 +53,14 @@ Adafruit_BNO055::Adafruit_BNO055(int32_t sensorID, uint8_t address)
...
 
+
 /**************************************************************************/
 /*!
-    @brief  Sets up the HW
+    Sets up the IMU Sensor for KBox, including axis- and sign mapping
 */
 /**************************************************************************/
-bool Adafruit_BNO055::begin(adafruit_bno055_opmode_t mode)
+bool Adafruit_BNO055::begin(adafruit_bno055_opmode_t mode, uint8_t axis_remap_orientation, uint8_t axis_remap_sign)
```
sarfata: 👍

lib/Adafruit_BNO055/Adafruit_BNO055.cpp
```diff
@@ -90,8 +88,7 @@ bool Adafruit_BNO055::begin(adafruit_bno055_opmode_t mode)
     //delay(100);
   //}
   //delay(50);
-
-  DEBUG("rebooted");
+  //DEBUG("rebooted");
```
sarfata: we should just remove this I think.

src/host/config/KBoxConfigParser.cpp
```diff
@@ -54,6 +54,11 @@ void KBoxConfigParser::defaultConfig(KBoxConfig &config) {
 
   config.imuConfig.enabled = true;
   config.imuConfig.frequency = 20;
+  config.imuConfig.enabled = true;           // enable internal IMU sensor
+  config.imuConfig.enableHdg = true;         // true if values taken from internal sensor
+  config.imuConfig.enableHeelPitch = true;   // true if values taken from internal sensor
+  config.imuConfig.calHdg = 3;               // hdg valid, if calibration value greater equal
```
sarfata: I am still not convinced we need this. Besides you, and to a lesser extent me, nobody is going to understand what those values mean. 

If I agree to make 3 and 2 the defaults, are you ok removing this? ;)

src/host/main.cpp
```diff
@@ -134,6 +135,12 @@ void setup() {
   BatteryMonitorPage *batPage = new BatteryMonitorPage(skHub);
   mfd.addPage(batPage);
 
+  if (config.imuConfig.enabled) {
```
sarfata: 👍 great idea.

src/host/pages/IMUMonitorPage.cpp
```diff
@@ -0,0 +1,102 @@
...
+  addLayer(_hdgTL);
+  addLayer(_calTL);
+  addLayer(_rollTL);
+  addLayer(_pitchTL);
+
+  hub.subscribe(this);
+}
+
+bool IMUMonitorPage::processEvent(const ButtonEvent &be){
+  // DEBUG("EventTypeButton: %i", be.clickType);
```
sarfata: We need to come up with a better way to disable logging for just one class. Commenting DEBUG is really not ideal, and removing them is also not great. I need to come up with something!

src/host/pages/IMUMonitorPage.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+  // TODO: Some damping for the display
+  _hdgTL->setText(String( SKRadToDeg(_heading), 1) + "°        ");
+  _calTL->setText(String( _magCalibration) + "/" + String( _accelCalibration) + "   ");
+  _pitchTL->setText(String( SKRadToDeg(_pitch), 1) + "°     ");
+  _rollTL->setText(String( SKRadToDeg(_roll), 1) + "°     ");
+
+  // Always show Hdg from IMU-sensor, but if the value is not trusted (which means
+  // calibration below config setting, change color to red
+	if ((_magCalibration <= _config.calHdg)||(_accelCalibration <= _config.calHeelPitch)) {
```
sarfata: nitpick: indentation problem here

src/host/pages/IMUMonitorPage.h
```diff
@@ -0,0 +1,53 @@
...
+#include "common/ui/Page.h"
+#include "common/ui/TextLayer.h"
+#include "common/signalk/SKHub.h"
+#include "common/signalk/SKSubscriber.h"
+#include "services/IMUService.h"
+#include "host/config/IMUConfig.h"
+
+class IMUMonitorPage : public Page, public SKSubscriber {
+	private:
+		TextLayer *_hdgTL, *_rollTL, *_pitchTL, *_calTL;
```
sarfata: indent issues here too (tabs instead of spaces)

src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,125 @@
...
-  if (sysCalib == 3) {
-    update.setNavigationAttitude(SKTypeAttitude(/* roll */ SKDegToRad(roll), /* pitch */ SKDegToRad(pitch), /* yaw */ 0));
-    update.setNavigationHeadingMagnetic(SKDegToRad(heading));
+  // update NavigationHeadingMagnetic if quality (= calibration value) is ok
+  if (_magCalib >= _config.calHdg) {
+    update.setNavigationHeadingMagnetic(_heading);
     _skHub.publish(update);
   }
+  // update NavigationAttitude if quality (= calibration value) is ok
+	if (_accelCalib >= _config.calHeelPitch && _gyroCalib >= _config.calHeelPitch) {
```
sarfata: nitpick: indent issue here again

src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,125 @@
...
+    //DEBUG("Heel = %.3f | Pitch = %.3f | Heading = %.3f", SKRadToDeg( _roll + _offsetRoll), SKRadToDeg( _pitch + _offsetPitch), SKRadToDeg(_heading));
+    _skHub.publish(update);
+  }
+}
+
+// get the actual values of calibration, heel & pitch (including offset) and heading for display
+// all angles in radians
+void IMUService::getLastValues(int &accelCalibration, double &pitch, double &roll, int &magCalibration, double &heading){
+  accelCalibration = _accelCalib;
+  pitch = _pitch + _offsetPitch;
```
sarfata: is it possible that the offset pushes the value beyond a theoretical maximum?

For example max pitch is +180, if the boat is pointing down and the offset is 5 degrees, we could have 179+5=184 which would be outside our range. Maybe we should apply some normalization here.

src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,125 @@
...
   // Heading and the Course Over Ground Magnetic.
 
-  /* if orientation == MOUNTED_ON_PORT_HULL */
-  double roll, pitch, heading;
-  roll = eulerAngles.z();
-  pitch = eulerAngles.y();
-  heading = fmod(eulerAngles.x() + 270, 360);
+  // TODO: different calcs according to mounting position
+  /* if orientation == MOUNTED_ON_STARBOARD_HULL */
+  _roll = SKDegToRad(eulerAngles.z());
```
sarfata: why dont you apply the offset here?

src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,125 @@
...
+}
+
+//  Calc offset for making Heel=0 and Pitch=0 (e.g. called with long button press in IMUMonitorPage)
+void IMUService::setOffset() {
+  // FIXIT: Why is it called at startup of KBox? event longClick coming?
+  if ( millis() > 10000 ) {
+    _offsetRoll = _roll * (-1);
+    _offsetPitch = _pitch * (-1);
+    INFO("Offset changed: Heel -> %.3f | Pitch -> %.3f", SKRadToDeg(_offsetRoll), SKRadToDeg(_offsetPitch));
+    // TODO: angle of heel and pitch must be smaller than 90°, check overflow!
```
sarfata: yes this is what i meant in my comment above :) I see you thought about this already.

src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,125 @@
...
+  bno055.getSensorOffsets(newCalib);
+
+  DEBUG("\n\nStoring calibration data to EEPROM...");
+  bno055.getSensor(&sensor);
+  bnoID = sensor.sensor_id;
+
+  EEPROM.put(_eeAddress, bnoID);
+
+  _eeAddress += sizeof(long);
+  EEPROM.put(_eeAddress, newCalib);
```
sarfata: Have you tried using the `src/host/util/PersistentStorage.cpp`? Do you want me to help with this?

src/host/pages/IMUMonitorPage.cpp
sarfata: I think the IMUService should expose a `isMagCalibrated()` and is `isHeelAndPitchCalibrated()` methods. This way the actual technique we use to decide if we are calibrated is not duplicated in two completely different part of the code.

src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,125 @@
...
+  // update NavigationHeadingMagnetic if quality (= calibration value) is ok
+  if (_magCalib >= _config.calHdg) {
+    update.setNavigationHeadingMagnetic(_heading);
     _skHub.publish(update);
   }
+  // update NavigationAttitude if quality (= calibration value) is ok
+	if (_accelCalib >= _config.calHeelPitch && _gyroCalib >= _config.calHeelPitch) {
+    update.setNavigationAttitude(SKTypeAttitude(/* roll */ _roll + _offsetRoll, /* pitch */ _pitch + _offsetPitch, /* yaw */ 0));
+    //DEBUG("Heel = %.3f | Pitch = %.3f | Heading = %.3f", SKRadToDeg( _roll + _offsetRoll), SKRadToDeg( _pitch + _offsetPitch), SKRadToDeg(_heading));
+    _skHub.publish(update);
```
sarfata: You should publish only once. Not twice.

With the current code, if both are calibrated you will send one update with heading first and then one update with heading + attitude. I do not think this is what you want.

sarfata:
Also let me know when you want to merge. I think besides a few details/nitpicks I mentioned above we could merge as-is and do more PRs for the rest.

lib/Adafruit_BNO055/Adafruit_BNO055.cpp
ronzeiller: If you do not need your remarks anymore, I can delete them.
But added you as collaborator, so you can make modifications too, if you want.

src/host/config/KBoxConfigParser.cpp
ronzeiller: :-))
yes, this is ok.
As I can see the heading on the IMUPage now (independently from the SKUpdate, which means independently from any cal-value) it is ok for me.

src/host/pages/IMUMonitorPage.cpp
ronzeiller: A lot of this DEBUG I just need during inventing of code. Then, when I think it works as I expect, I think it is a pity to delete them....

src/host/pages/IMUMonitorPage.cpp
ronzeiller: I do not know how to get rid of ident-problem. In every editor I have it shows a 2-ident tab.
In Github it shows wrong. If I correct 8 -> 2 in github I can not save because it is said "no changes"

src/host/services/IMUService.cpp
ronzeiller: I need the original but "turned for mounting" values to calculate an offset.
When I apply the offset here it works just for the first time the button is pressed.
If the button is pressed again, then an offset from an offset would be calculated.

src/host/services/IMUService.cpp
ronzeiller: The offset is meant for doing small corrections only, not for correcting mounting positions. E.g. after loading people, water, luggage.
If pitch is around 180 then you have (other) problems :-)
 If someone is sailing with heel 40° and wants to do have heel 0 on the display then this would be crazy.

src/host/services/IMUService.cpp
ronzeiller: So, yes, in theory the value could be pushed to an out of range value, but I do not know if normalization is really needed.....?

src/host/services/IMUService.cpp
ronzeiller: OK, then I did not understand the update.

can I write ```_skHub.publish(update);```
just once, regardless if there was ```update.setAttitude.....```?
(Like I have it now?)

src/host/services/IMUService.cpp
ronzeiller: To be honest, I do not understand the PersistentStorage up to now, or did not look at it for longer to really understand it. ;-)

I thought that the calibration values are stored in the MCU of the BNO055 directly, so it is not needed here anyway?

ronzeiller:
Hi Thomas,
yes, if there are no fundamental errors anymore, it would be nice if you could merge. 

As I have worked already on the positioning config this feature is already in this PR

Just to keep in mind:
1. delete config settings for cal including all tests
2. solve the ident problem
3. may be do the offset calculation before update
4. Change to PersistentStorage
5. Adafruit library delete uncommented code
6. Offset beyond maximum values

src/host/pages/IMUMonitorPage.cpp
sarfata: Re-Indent: this article might help: https://stackoverflow.com/questions/4119872/netbeans-removing-trailing-whitespace-on-save-and-tabs-to-spaces

I think your editor is configured to show tabs as 2 spaces. this is why it looks good to you. But since they are actually tabs, it does not look good in github (they do this on purpose to make it easier to detect).

I can push a commit to fix it if needed.

CaptainRon47:
All,

I am trying to follow what you are all doing and I thought I would (finally) try and do it using GitHub.  In the past I would just copy a zip file and open in my desktop with Atom.  So I created my own Repository (on the GitHub website) and imported RonZ’s latest updates (as of this morning, but changes still going on I see).  Then I cloned this into my desktop GitHub and opened it with Atom.  Now the question is, as changes are made by RonZ (for example: in experiment/chibios – less than a minute ago) will those changes show up in my Repository on the web and or on my desktop version, or is my clone frozen as it is?  

 

I guess I will find out shortly!

Happy New Year,

Ron

ronzeiller:
@CaptainRon47 CaptainRon47
Hi Ron,

Happy New Year to you too!

> I am trying to follow what you are all doing and I thought I would (finally) try and do it using GitHub.  In the past I would just copy a zip file and open in my desktop with Atom.  So I created my own Repository (on the GitHub website) and imported RonZ’s latest updates (as of this morning, but changes still going on I see). 

You should go to my github and do a fork -> then you get a copy of my work. It is then connected to my github which makes things easier for you to update.
Be aware, that my code is often just for discussion for new features or beta code.

I would strongly recommend to fork the original from @sarfata 
Also the good new config stuff you will get there!

After forking, you clone  again, to get a local working copy.

> Now the question is, as changes are made by RonZ (for example: in experiment/chibios – less than a minute ago) will those changes show up in my Repository on the web and or on my desktop version, or is my clone frozen as it is?

This is the status and it does not change automatically.
For update you can do something like this directly in the terminal of Atom:
(You have to do this only once)

```
git checkout master (....or the branch name you like to update)

git remote add upstream https://github.com/sarfata/kbox-firmware.git
```

Now you are connected to your own "github" and the original "sarfata-code" called "upstream" in this example.

This shows you the different  remote names:
`git remote –v`

For example I have here (github and upstream are just names, you could take different here):

```
github  https://github.com/ronzeiller/kbox-firmware.git (fetch)
github  https://github.com/ronzeiller/kbox-firmware.git (push)
upstream  https://github.com/sarfata/kbox-firmware.git (fetch)
upstream  https://github.com/sarfata/kbox-firmware.git (push)
```

**For updating master from sarfata I just do:**

```
git checkout master
git reset master to upstream/master       (master or the name of the branch I want to update....)
git fetch upstream
git reset --hard upstream/master
git push --force
```
Then my github and my local working copy is up-to-date again with all changes sarfata made.

Then the best is to create a new branch in Atom to work and develop, or with your special adjustments......
You can later always merge new things to this branch too.

For all of this and a lot more, you will find really good descriptions in google on github or stack exchange.

Good luck!

CaptainRon47:
Thanks for the helpful discussion.  I must admit I am intimidated by the unfamiliar terminology (pushes, pulls, forking, etc).  In principle I understand what is happening but not having lived with this open programming environment for long I tend to play it safe and just copy a zip file to my local machine and work with that.  But, as I said, I am trying to ‘get with it,’ and start to use GitHub as intended.  I did ‘import’ RonZ’s  IMU_Page version into my own Repository as a ‘new’ project and then cloned it to my pc.  I wanted to see the latest changes with the new display page and the NMEA outputs activated.  I will try your suggestion of ‘forking’ the repository at the next release from Thomas.

And so the **GOOD** news is that it all works!  I get the IMU page in the display!   I reconfigured the speeds on the NMEA1 and 2 inputs/outputs to match my equipment and I modified the IMU orientation to reflect my VerticalToptoBow mounting.  I changed these in KboxConfigParser.cpp since I am not sure if the idea of a config file on the SD card is working in this version.   I get the correct Roll and Pitch data on the display (a long push on the button does zero out the values) and also get NMEA data out on both NMEA1 and 2 at the correct speeds.  The Wifi data and the USB output all looks good.  I see that the current USB output is back to being a DEBUG output, which is fine now that the NMEA outputs are working, as I can use that to connect to my Rpi running OpenCPN instead of the USB.  I have been unable to test if SignalK data is available from Kbox as my Rpi is acting up!

Now the **NOT SO GOOD** news, my USB connector attachment came off of the Kbox board.  I probably am a bit more active, than most, in moving the Kbox from my boat to my office where I make programming changes.  In the office the USB cable (I have a long one) hangs off the desk to connect to the PC tower on the floor.  The weight of the cable and movement from boat to office was enough handling to loosen the connector on the board.  I leave the USB cable plugged into the Kbox all the time, but I did not think to build a strain relief to the plastic box.  I did drill a hole in that box so the USB cable could be plugged in while the board was in the box and I should have added strain relief as the pinout cables have.  I was able to solder the connector attachment back to the board and I at least get power, but I did not get all 5 tiny pins soldered back down so I don’t get a USB port for programming the Kbox.  My soldering skills and equipment are not up to that task.  

Thomas, 
Send me a private email and we can discuss what you want to do.

Thanks,
RonS

sarfata:
@CaptainRon47 I have sent you an email. We will get this fixed. Thanks for your time testing and playing with KBox!

Issue #91, opened 2017-12-10
Should not need dynamic memory in NMEA and NMEA2000 visitors


sarfata:
Right now the `SKNMEAVisitor` and `SKNMEA2000Visitor` use a lot of dynamic memory to generate messages. This is not a great solution. We should try to use as little dynamic memory as possible.

PR #94, opened 2017-12-10
Feature/refactor signalk conversions

sarfata:
This improves:
 - How we convert SignalK to NMEA2000 and should make it a lot more readable
   Instead of using a pseudo visitor pattern that was not really one in this case and was pretty hard to understand, I propose we have a `convert()` function that looks at the incoming updates (and soon it will look at the configuration too) and decides what to do. If the conversion if very simple we can do it directly in `convert()`, if it's more complicated we can use a private function to do the conversion.

 - Memory management during conversions
   The previous approach used dynamic memory allocation (`new`) to create the `tN2kMsg` objects and they were "returned" as reference to the caller who had to do something with them before destroying the `NMEA2000Visitor` who "owned" the memory.

    This has many issues:
   - It was way too easy to create object and leak them if you are not very careful
   - It leads to memory fragmentation because we dynamically allocate lots of small objects and then destroy them

    The new approach requires that the object who calls the `convert()` function must pass a reference to an object that implements the "NMEA2000ConversionOutput" interface. This interface has one method `pushMessage()` that will be called every time a message needs to be pushed. In the `NMEA2000Service` we can push the messages directly to the CAN driver which will buffer and send them.
    This completely eliminates the memory management issue, does not use any dynamic memory and seems like a much better approach to me.

Comments/Suggestions?


sarfata:
@ronzeiller I would love to know what you think about this change. Do you think it makes things easier to read?

ronzeiller:
It took me quite a while to (hardly) understand the Visitor Pattern, I think now it will be much easier to read the code and data signaling. Also the memory issue sounds great!
Also that NMEA0183 and N2k are similar now is nice.
And it will prepare the ground for different outputs and the data-computer if I understood it right.
:sunglasses: :sailboat:

sarfata:
This closes #91

Issue #13, opened 2016-09-09
Make it easy to customize essential parameters of the KBox


sarfata:
We need to make basic configuration much easier but it will be configured in a header file and will require recompiling the firmware.
- [x] Speed of NMEA1 and NMEA2
- [ ] Orientation of KBox in the boat
- [x] Enable / Disable IMU

sarfata:
closed by #107

Issue #71, opened 2017-11-23
As a user, I should be able to completely disable NMEA2000


sarfata:

CaptainRon47:
Probably setting a flag in the beginning somewhere to output NMEA 2K data or not would resolve this.

sarfata:
There is a configuration for this now.

closed by #107.

Issue #62, opened 2017-11-23
As a user, I want to configure the speed of NMEA input/output


sarfata:
Users should be able to configure the speed of NMEA ports without recompiling the firmware.

CaptainRon47:
Thomas,
I have my NMEA interface box connected to NMEA1 and it runs at 4800, while I have my radio AIS output connected to NMEA2 and it runs at 38400.  So I have them set as such.  Once they are set I don't see a need to change them, so the request to set them without recompiling is not mine.

sarfata:
Thanks for the precision Ron. I do think that this is one of the most basic configuration and people should not have to recompile for that so I hope to have a configuration file on the sdcard that defines this.

sarfata:
closed by #107

Issue #43, opened 2017-08-29
As a user, I would like to see NMEA data going out the NMEA outputs


ronzeiller:
You write:
Two NMEA0183 outputs

How to activate NMEA0183 output?

sarfata:
Hi,

They are not used in the code yet but you can just write to Serial2 and Serial3. What information would you like to see on those outputs? Tell me more and I will help!

thomas

ronzeiller:
Hi Thomas,

All informations available, which can be described as pure NMEA0183 could be written to NMEA1 or NMEA2. 
(The $PCDIN sentences would probably make no sense on Serial Output, as not understood from any navigation program directly connected)
This would make KBox more universatile in its use.
For examle is this needed to connect KBox to a Raymarine´s (formerly Tacktick) Micronet Interface Transmitter for communication with Racemaster and/or other Micronet Displays.

Thank you!
Ronnie

CaptainRon47:
My specific interest here is to be able to set way points from OpenCPN and send them to my GPS.  But my GPS only has one input (at 4800) and so I cannot send it both input instrument data and waypoint data separately.  Currently my GPS, at the helm, receives all my instrument NMEA data from my NMEA instrument interface box as it is also sent to the KBox (my NMEA interface box can source to both the GPS and Kbox simultaneously).  I would like to be able to (instead) repeat all my instrument data and add way point sentences from OpenCPN (on my Rpi) as it returns to the Kbox and comes from the NMEA outputs on the Kbox.  So KBox would have to act like a true multiplexer by both repeating the input data and adding data coming from OpenCPN (or any other navigation program).

sarfata:
I think we can make a dumb version of output to NMEA work easily but we are going to run into issues when there is too much data flowing in KBox than what we can send at 4800 bauds. In this case, some messages will get lost and we will need some kind of filtering, or reduction of speed to make sure this does not happen. We can implement this later.

sarfata:
I think we can make a dumb version of output to NMEA work easily but we are going to run into issues when there is more data flowing in KBox than we can send at 4800 bauds. In this case, some messages will get lost and we will need some kind of filtering, or reduction of speed to make sure this does not happen. We can implement this later.

CaptainRon47:
Thomas,
[...]

PR #104, opened 2017-12-14
#13 - KBox configuration

sarfata:
Very much a work in progress. There is just unit tests for now. I still need to plug it into the nmea and imu services.

Sharing so you can give me your opinion and thoughts @ronzeiller .


ronzeiller:
I think I will have to get used to this JSON thing :-)

First view: looks good :thumbsup:
Second view: may be you could rename NMEA_1, NMEAConfig and all sorts of this to something which describes it better, e.g. COM1 and COM2 (as nice computers still have :grin: )
There could be so much more on these interfaces as Seatalk, PCDIN-sentences, input of extern IMU-sensor or other sensors in a binary format or Actisense format output as is recognised by Expedition software and so on....

Then NMEA-naming will belong to special things as Talker-IDs, sentence-IDs, enabling output of sentences....

ronzeiller:
@sarfata
Thought more about this config issue...

I am afraid that this kind of config will be very, very complicated over all.
We will have around 50 to 100 (or even more) different config settings in different classes where somebody has to search where to find what, because it is not clear where a config variable will belong to.

Then this settings will be boolean, strings, arrays or multidimensional arrays and for all of this tests, and functions will be needed.

I know, that my ini-file with it's simple structure is from programers point of view quite ugly, but it is very simple and fast to implement.

I would suggest:

1.  that we first make a collection of variables already needed, all in one class, very easy to overlook and easy to use all over the code.

2. Then complete all needed functions, inputs and outputs on serial and web interfaces that KBox can be practically used on a boat.

3. Then let´s start to work on the data computer (lot's of more settings, constants, boat parameters, calibration values,.......)

4. And then, at least, when all of this usefull things are working, start a web based configuration thing with this kind of classes, JSON, get set.....

Just my 2cents....

sarfata:
> Second view: may be you could rename NMEA_1, NMEAConfig and all sorts of this to something which describes it better, e.g. COM1 and COM2 (as nice computers still have 😁 )

That is a very good point and I have renamed the config and the service to "SerialService" which should be more generic.

> I am afraid that this kind of config will be very, very complicated over all.
> We will have around 50 to 100 (or even more) different config settings in different classes where somebody has to search where to find what, because it is not clear where a config variable will belong to.

The idea with my system is that each service has one associated configuration class. The reason for doing this is it limits interdependencies between services. The SerialService for example, only knows about SerialConfig, not about the config required for other services.

I hope it will not be too complicated but I understand your concerns and so I have decided to push this PR a bit further and implement a few more config variables, and plug them in the code to see what it looks like. Let's re-discuss when I am done and see if it's too verbose for what it gives us.

> Then this settings will be boolean, strings, arrays or multidimensional arrays and for all of this tests, and functions will be needed.

If we are reading from a INI file we still have to convert to bools, ints, enums or arrays, no? And if we are doing this type of parsing, I think we still need unit tests. I am not sure the choice of JSON or INI changes anything here. Can you explain why INI makes this easier?

ronzeiller:
> If we are reading from a INI file we still have to convert to bools, ints, enums or arrays, no? And if we are doing this type of parsing, I think we still need unit tests. I am not sure the choice of JSON or INI changes anything here. Can you explain why INI makes this easier?

What I meant is, that for implementing new values it is very easy, just write into the according section of **one central** file `xy = nn;` or `xy = array{a,b,c}` and ready you are :)

> The idea with my system is that each service has one associated configuration class. The reason for doing this is it limits interdependencies between services. The SerialService for example, only knows about SerialConfig, not about the config required for other services.

I see your point.
Just thinking of the future where lot of inputs/outputs and data formats will be implemented that it could be that too much encapsulation will make it difficult to read code. and find where, what is.
But as you said: Let's re-discuss it when it becomes more visible where it is going to.

ronzeiller:
Apropos easy.... :-)

Could you please do the special string thing into the Generator for the SignalK Time variable, whenever you find time for this? [85#issueComment](https://github.com/sarfata/kbox-firmware/issues/85#issuecomment-348402690)
Then I could start with the KBox time and the timestamp for logging files....

sarfata:
@ronzeiller would you mind taking one more look and let me know if you see anything wrong?

I will continue working on configuring the WiFi mode, SSID and Password in a separate PR.

ronzeiller:
@sarfata 
This is a huge and great step forward! 
So glad that you took the efford of doing this.
I cannot see anything wrong.
 :thumbsup:  :thumbsup:

sarfata:
closes #13 
closes #71 
closes #62 
closes #43

PR #105, opened 2017-12-18
PGN127275 Heel, Pitch, Yaw added + PGN128267 Water Depth discussion

ronzeiller:
**Question/TODO:** 
check if NaN can be stored into SKUpdate for Attitude. The background is, what happens if one of the values pitch, roll, yaw is not delivered by the N2k bus?

Same issue with internal sensor, where it is written yaw = 0 which is not correct?

And we need a config setting for reading internal sensor or external attitude values from bus
TODO: if switched to read from N2k bus than the IMUPage has to be switched too.


ronzeiller:
We will have to check yaw separately as according to Maretron SSC200
Solid State Rate/Gyro Compass Manual  for example:

**PGN 127257 – Attitude**
The SSC200 uses this PGN to indicate the vessel’s attitude (pitch and roll). **The Yaw (field 2)
is not used, therefore this field always contains 0x7FFF (data not available).** 

As obviously there is a difference between zero and data not available this could be an issue for SignalK? (How is handled NaN there?)

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -208,6 +208,38 @@ const SKUpdate& SKNMEA2000Parser::parse127250(const SKSourceInput& input, const
...
+  double pitch = N2kDoubleNA;
+  double roll  = N2kDoubleNA;
+
+  if (ParseN2kPGN127257(msg, sid, yaw, pitch, roll)) {
+    SKUpdateStatic<1> *update = new SKUpdateStatic<1>();
+    update->setTimestamp(timestamp);
+
+    SKSource source = SKSource::sourceForNMEA2000(input, msg.PGN, msg.Priority, msg.Source);
+    update->setSource(source);
+    // TODO: check if it is possible to set NAN as value to SKUpdate
```
sarfata: Setting NAN should work and seems like the best solution to me. We will need to make sure that when we serialize the data to JSON we skip fields that have the value NAN.

src/test/signalk/SKNMEA2000ParserTest.cpp
```diff
@@ -111,6 +111,16 @@ TEST_CASE("SKNMEA2000Parser: Basic tests") {
...
 
+  SECTION("127275: NavigationAttitude Heel, Pitch, Yaw") {
+    // SID, yaw, pitch, roll in radians
+    SetN2kAttitude(msg, 0, SKDegToRad(179.9), SKDegToRad(-2.3), SKDegToRad(3.2));
+    const SKUpdate &update = p.parse(SKSourceInputNMEA2000, msg, SKTime(0));
+    CHECK( update.getSize() == 1);
+    CHECK( update.getNavigationAttitude().yaw   == Approx(SKDegToRad(179.9)).epsilon(0.0001) );
+    CHECK( update.getNavigationAttitude().pitch == Approx(SKDegToRad(-2.3)).epsilon(0.0001) );
+    CHECK( update.getNavigationAttitude().roll  == Approx(SKDegToRad(3.2)).epsilon(0.0001) );
+  }
```
sarfata: We should had a check with NAN as the Yaw value here. Just to confirm it gives what we want.

ronzeiller:
@sarfata 
Hi Thomas,
heel/pitch/yaw is working now as expected with NaN

**Also did some corrections on Water Depth:**
Reading some papers I came to the conclusion that water depth is a little bit more complicated. 
It seems to be either depth below surface or depth below lowest point.
So I changed the code and tests accordingly.

See for example:
https://www.maretron.com/support/manuals/DST110UM_1.3.html
https://seatronic.no/productattachments/index/download?id=144

The test fails because of 1.9 == 1.9 ?????

Also the test will fail if offset is e.g. 1.995 but as stated in the DST110 Manual offset may be 1.e-3

Looking forward what you say about this.

ronzeiller:
...just saw that git makes it very complicate to read the changes because I also sorted the tests according the PGN.
Sorry for that, next time I will do the changes in code first and do the sorting later, when code is approved.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -244,32 +272,56 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
 
-      update->setEnvironmentDepthBelowTransducer(depthBelowTransducer);
-
-      // When offset is negative, it's the distance between transducer and keel
       if (!N2kIsNA(offset)) {
-        if (offset < 0) {
+        if (offset == 0) {
+          // When you have zero offset than the transducer is mounted on the
+          // lowest point of the boat, e.g. motor boat w/o keel, sailing boat keel
+          update->setEnvironmentDepthSurfaceToTransducer(N2kDoubleNA);
```
sarfata: I do not think we need to do that here. In that case we can just omit these two variables. This is what the SignalK official tools do.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -244,32 +272,56 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
+          update->setEnvironmentDepthBelowSurface(N2kDoubleNA);
+          update->setEnvironmentDepthBelowKeel(depth);
+          update->setEnvironmentDepthTransducerToKeel(0);
+        }
+        else if (offset < 0) {
+          // When offset is negative, it's the distance between transducer and keel
+          update->setEnvironmentDepthSurfaceToTransducer(N2kDoubleNA);
```
sarfata: Same here, we can just ignore the field we do not have. It's better to not include them than to include them with value NA.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -244,32 +272,56 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
         }
         else if (offset > 0) {
+          //  Positive offset numbers provide the distance from the transducer to the waterline.
           update->setEnvironmentDepthSurfaceToTransducer(offset);
-          update->setEnvironmentDepthBelowSurface(depthBelowTransducer + offset);
+          update->setEnvironmentDepthBelowSurface(depth + offset);
+          update->setEnvironmentDepthBelowKeel(N2kDoubleNA);
+          update->setEnvironmentDepthTransducerToKeel(N2kDoubleNA);
         }
+      } else {
```
sarfata: If you are not setting the NA values, then you do not need this.

src/common/signalk/SKNMEA2000Parser.cpp
```diff
@@ -244,32 +272,56 @@ const SKUpdate& SKNMEA2000Parser::parse128259(const SKSourceInput& input, const
...
       update->setSource(source);
+      update->setEnvironmentDepthBelowTransducer(depth);
 
-      update->setEnvironmentDepthBelowTransducer(depthBelowTransducer);
-
-      // When offset is negative, it's the distance between transducer and keel
       if (!N2kIsNA(offset)) {
-        if (offset < 0) {
+        if (offset == 0) {
+          // When you have zero offset than the transducer is mounted on the
```
sarfata: Did you find a source for this information?

src/sktool/tests/n2k-transducer.json
```diff
@@ -19,6 +19,8 @@
       { "path": "environment.depth.belowTransducer", "value": 3.45 },
       { "path": "environment.depth.belowSurface", "value": 4.364 },
       { "path": "environment.depth.surfaceToTransducer", "value": 0.914 }
+      { "path": "environment.depth.transducerToKeel", "value": -1000000000.0 }
```
sarfata: No. I don't think this is correct. We should not generate these values for this incoming message. 

The SignalK tools do not do it, and I would bet that iKommunicate does not do it either.

src/common/signalk/SKNMEA2000Parser.cpp
sarfata: The Maretron manual says:

> 3:  Offset – If the value of this field is positive, it represents the difference in depth between the transducer and the waterline of the vessel.  If the value of this field is negative, it represents the difference in depth between the transducer and the keel of the vessel.  The unit of measurement for this field is 1x10-3 m.

I think when the value is 0 we do not know where it is.

sarfata:
> ...just saw that git makes it very complicate to read the changes because I also sorted the tests according the PGN.
> Sorry for that, next time I will do the changes in code first and do the sorting later, when code is approved.

👍  Also could have split the depth changes to another PR just to keep things separate and easier to discuss.

src/common/signalk/SKNMEA2000Parser.cpp
sarfata: Also the Raymarine manual says:
> If an offset is not applied, displayed depth readings represent the distance from the transducer to the sea bed

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: The reason why I actively set NaN was that I was not sure if it is ok 
to set SKUpdateStatic<3> and then just 1 value will be stored in update.
(I think I had an issue with a test, but may be I mix something up.....)

If you say that does not matter, then I drop all NaN and the else of course.

src/sktool/tests/n2k-transducer.json
ronzeiller: May be, but they could be wrong... :joy:

There was an error in the test, which was gone after I entered this....

src/common/signalk/SKNMEA2000Parser.cpp
ronzeiller: A mayor difference to this we had before is, that offset = zero is a defined state 

> I think when the value is 0 we do not know where it is.

Generally spoken: you never know (by the instruments) the depth from the keel to the ground, when set to "depth from water surface" and on opposite you do not know where the transducer is mounted if set to "depth below keel".

**For offset=0**
Think of a boat, where the transducer **is** at the lowest point.
Then you would enter zero to offset if you want to display depth below keel, no?
If an offset is zero, than **we do know that someone had set it on purpose.** 
(Otherwise it would be NaN)

If someone does not enter any offset, then we know that the device was not set up (correctly). So we just can say it is the depth below sensor and not more.
This was and is still the way the code is doing it.
I think that this is meant with "not applied" (as zero is an active setting)

But may be I am totally wrong and the offset is Zero on a brand new unit.
We should ask somebody from Raymarine, Maretron or Airmar if they make a difference between "not set" / "not applied" and defined 0......

And what will then  actually be displayed in the two different settings **depth below keel** and **depth below water surface**???

src/sktool/tests/n2k-transducer.json
ronzeiller: Took it out again

sarfata: 😧 There is absolutely no way that `-1000000000.0` is correct here :no_good: 

Either we generate: `"value": null` or we do not generate an entry for `transducerToKeel` at all. In general SignalK omits values that are `null` or `undefined`. This is what we should do here. I believe it is correct.

> There was an error in the test, which was gone after I entered this....

or maybe the test was failing because the code is incorrect 😜

src/common/signalk/SKNMEA2000Parser.cpp
sarfata: Yes it is ok.

src/common/signalk/SKNMEA2000Parser.cpp
sarfata: > Think of a boat, where the transducer is at the lowest point.

Yes but think of a boat where the transducer is right at the floatation point. 😛 

I think they gave up on these two situations when they decided to use a positive/negative value in the offset to indicate the reference. They just do not support it and I think we should not support it either.

I think in practice, you would say `+0.01` if it's right below the keel and `-0.01` if it's right below the surface. Saying 0 is like saying no offset and in that case we do not know where the sensor is located.

> But may be I am totally wrong and the offset is Zero on a brand new unit.

Yes it would be useful to know what a new device sends.

> And what will then actually be displayed in the two different settings depth below keel and depth below water surface???

If we do not have the information we just do not send it in SignalK. Displays should not show it either. That is a very common thing. There are lots of possible data on the network and it's not uncommon to be missing data.

ronzeiller:
I think you are right and you are definitely right when you say that we should not interpret those kind of navigational datas.
I took out all changes from Water Depth and the transducer.json and put back to original. If we get new informations we have at least discussed some of it already :-)

One test is added with offset = 0, may be you like it.

So we are back to the heel/pitch/yaw...
On the internal IMU sensor you put 0 to yaw, because it is not present. IMHO this is not so good, because in fact it is not zero. (But for me personally it doesn't mean a big thing, because I will not rely on the yaw anyway and will just take heel and pitch)

In the parser it is different, because Timo's library will send NaN if not present on the bus.
The test is working fine.

> One thing we should do though is test what the output looks like for the yaw when it is NAN. Can you add a test in `src/sktool/tests` for this? I believe the proper thing to do is to not include the value for yaw in the output.

I am sorry, but I do not understand this.
Where did you get the PCDIN sentences from?  :confused:

sarfata:
> On the internal IMU sensor you put 0 to yaw, because it is not present. IMHO this is not so good, because in fact it is not zero. (But for me personally it doesn't mean a big thing, because I will not rely on the yaw anyway and will just take heel and pitch)

I completely agree. This is not good. We should save NAN in the SKUpdate and make sure that when we generate the JSON, we do not print the field "yaw" if the value is NAN. 

Let me know if you need help for this.

>>One thing we should do though is test what the output looks like for the yaw when it is NAN

> I am sorry, but I do not understand this. Where did you get the PCDIN sentences from? 😕

I used a log from my own boat but you could also use the NMEA2k simulator by Timo and record what it generated in $PCDIN format.

ronzeiller:
> > On the internal IMU sensor you put 0 to yaw, because it is not present. IMHO this is not so good, because in fact it is not zero. (But for me personally it doesn't mean a big thing, because I will not rely on the yaw anyway and will just take heel and pitch)

> I completely agree. This is not good. We should save NAN in the SKUpdate 

Just to do some documentation:

Timo says, that we can write NaN into the PGN's but have to take care which one -> [86#issuecomment](https://github.com/ttlappalainen/NMEA2000/issues/86#issuecomment-353285806)

In Signal K they have `NULL` for invalid data in json:
https://github.com/SignalK/specification/blob/c446914b331084dc5c8bcddf6aa65003b69b1c45/gitbook-docs/data_model.md#missing-or-invalid-data

Summary:
As we do not have `NULL` in C++ we stick to `N2kUInt16NA, N2kDoubleNA, N2kUInt32NA`
for N2k and SKUpdate, leave fields empty in NMEA0183 and JSON

> > We should save NAN in the SKUpdate and make sure that when we generate the JSON, we do not print the field "yaw" if the value is NAN.

> Let me know if you need help for this.

You mean in `SKJSONVisitor.cpp`, e.g. here?

```
case SKValue::SKValueTypeAttitude:
        {
          JsonObject &attitude = obj.createNestedObject("value");
          attitude["pitch"] = v.getAttitudeValue().pitch;
          attitude["roll"] = v.getAttitudeValue().roll;
          attitude["yaw"] = v.getAttitudeValue().yaw;
        }
        break;
case SKValue::SKValueTypePosition:
        JsonObject &position = obj.createNestedObject("value");
        position["latitude"] = v.getPositionValue().latitude;
        position["longitude"] = v.getPositionValue().longitude;
        position["altitude"] = v.getPositionValue().altitude;
        break;
```

But to be consequent, shouldn't that be made with **all** values then (also with depth, latitude, longitude, .......)?

And also in `SKNMEAConverter.cpp` like here:
```
if (update.hasNavigationAttitude()) {
....
 sb.setField(2, SKRadToDeg(update.getNavigationAttitude().pitch), 1);
```

where it should become something like:

```
if (update.hasNavigationAttitude()) { 
.....
if (update.getNavigationAttitude().pitch == N2kDoubleNA) {
   sb.setField(2, '');
} else {
   sb.setField(2, SKRadToDeg(update.getNavigationAttitude().pitch), 1);
}
```

In a first step this all for "self made" values from internal sensors?

But how are incoming "not valid"-values from N2k handled?

ronzeiller:
What do you think of declaring a const in `SKUnits.h `like:

`const double   doubleNaN=-1e9;`

for invalid or not available N2k values, so we do not need to `#include N2kMsg.h` everywhere?

Or do you want to declare it local where needed?

sarfata:
> What do you think of declaring a const in SKUnits.hlike:
> 
>     const double doubleNaN=-1e9;
> 
> for invalid or not available N2k values, so we do not need to #include N2kMsg.h everywhere?
> 
> Or do you want to declare it local where needed?

💯 Great idea. Maybe we call it SKDoubleNAN to avoid conflict with other libraries?

sarfata:
> **Summary:**
> As we do not have `NULL` in C++ we stick to `N2kDoubleNA`
> for N2k and SKUpdate and leave fields empty in NMEA0183 and JSON

Yes. Agree with you on this!

We could always use `null` sometimes but I have not seen other signalk implementation do this at the moment. As you pointed out, the spec authorizes it but is not really clear on when to use it.
 
> > > We should save NAN in the SKUpdate and make sure that when we generate the JSON, we do not print the field "yaw" if the value is NAN.
> 
> > Let me know if you need help for this.
> 
> You mean in `SKJSONVisitor.cpp`, e.g. here?

Yes.

> ```
> case SKValue::SKValueTypeAttitude:
>         {
>           JsonObject &attitude = obj.createNestedObject("value");
>           attitude["pitch"] = v.getAttitudeValue().pitch;
>           attitude["roll"] = v.getAttitudeValue().roll;
>           if ( v.getAttitudeValue().yaw != N2kDoubleNA)
>               attitude["yaw"] = v.getAttitudeValue().yaw;
>         }
>         break;
> case SKValue::SKValueTypePosition:
>         JsonObject &position = obj.createNestedObject("value");
>         position["latitude"] = v.getPositionValue().latitude;
>         position["longitude"] = v.getPositionValue().longitude;
>         position["altitude"] = v.getPositionValue().altitude;
>         break;
> ```

👍 

> But to be consequent, shouldn't that be made with **all** values then (also with pitch, heel, depth, latitude, longitude, .......)?

That is a great question. I think we may want to do it for all fields of composite values (like attitude or position). For example, we do not do it yet for position.altitude and I think it's incorrect. We should not report 0 for altitude when we do not know. We should just not print it.

> And also in `SKNMEAConverter.cpp` like here:
> ```
> if (update.hasNavigationAttitude()) {
> ....
>  sb.setField(2, SKRadToDeg(update.getNavigationAttitude().pitch), 1);
> ```
> 
> where it should become something like:
> 
> ```
> if (update.hasNavigationAttitude()) { 
> .....
> if (update.getNavigationAttitude().pitch == N2kDoubleNA) {
>    sb.setField(2, '');
> } else {
>    sb.setField(2, SKRadToDeg(update.getNavigationAttitude().pitch), 1);
> }
> ```

Absolutely. I am not sure what we would generate right now if the value is NAN but it would not be pretty probably ;)

ronzeiller:
Some NAN checks added in SKJSONVisitor and SKNMEAConverter

- don't know if it is necessary or does it hurt  to check NAN at Batteries and Pressure as is checked in the according Service? 
(As soon as those values are added to SKNMEA2000Parser it should be checked in the parser anyway)
- Not for Lat and Lon as should/could be checked at incoming PGN, but added to altitude because this could be NAN despite values for Lat/Lon are valid, or should they be checked here too?
- Rudder is checked in the NMEA2000Parser and in SKNMEAConverter

Some values (e.g.: True Wind datas) could get NAN in future "Data Computer" so may be it is safer to check them all again  in SKNMEAConverter?

ronzeiller:
@sarfata 
NAN checks added in SKJSONVisitor and SKNMEAConverter as discussed.
All changes requested should be implemented.
Should be ready for merging now.

sarfata:
@ronzeiller it looks like I do not have permissions to push a commit to this branch of your project.

I will make a PR against your project with my changes. You can merge it if you agree and then I will merge this PR.

ronzeiller:
:boat:

sarfata:
⛵️

PR #106, opened 2017-12-22
Feature/36 configuration

ronzeiller:
Hi Thomas,
could you please add something like this, then my T36-KBox is able to write and read from sd-card?

image


sarfata:
👍

PR #107, opened 2017-12-26
some more IMU config settings discussion

ronzeiller:
- Some more IMU config values needed for IMUMonitorPage, writing back to N2k bus and writing NMEA output for internal sensor, (if master nmea sentence is enabled)
- added tests 
- switched one of your false values to true to make it more like default KBox setup? (as we have now false for other values to test)

### DIFF
--- a/src/host/config/KBoxConfigParser.cpp
+++ b/src/host/config/KBoxConfigParser.cpp
@@ -46,7 +46,15 @@ void KBoxConfigParser::defaultConfig(KBoxConfig &config) {
   config.serial2Config.inputMode = SerialModeNMEA;
   config.serial2Config.outputMode = SerialModeDisabled;
 
-  config.imuConfig.enabled = true;
+  config.imuConfig.enabled = true;            // enable internal IMU sensor
+  config.imuConfig.enabledHdg = true;         // true if values taken from internal sensor
+  config.imuConfig.enabledHeelPitch = true;   // true if values taken from internal sensor
+  config.imuConfig.calHdg = 2;                // hdg valid, if calibration value greater equal
+  config.imuConfig.calHeelPitch = 2;          // heel (roll) & pitch valid, if calibration value greater equal
+  config.imuConfig.writeN2kHdg = false;       // write heading to N2k bus
+  config.imuConfig.writeN2kHeelPitch = false; // write heel & pitch values to N2k bus
+  config.imuConfig.writeNMEAHdg = true;       // if NMEA hdm enabled write value from unternal sensor
+  config.imuConfig.writeNMEAHeelPitch = true; // if NMEA xdrAttitude enabled, write value from internal sensor
 }
 
 void KBoxConfigParser::parseKBoxConfig(const JsonObject &json, KBoxConfig &config) {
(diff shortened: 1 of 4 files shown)


src/host/config/KBoxConfigParser.cpp
```diff
@@ -46,7 +46,15 @@ void KBoxConfigParser::defaultConfig(KBoxConfig &config) {
   config.serial2Config.inputMode = SerialModeNMEA;
   config.serial2Config.outputMode = SerialModeDisabled;
 
-  config.imuConfig.enabled = true;
+  config.imuConfig.enabled = true;            // enable internal IMU sensor
+  config.imuConfig.enabledHdg = true;         // true if values taken from internal sensor
+  config.imuConfig.enabledHeelPitch = true;   // true if values taken from internal sensor
+  config.imuConfig.calHdg = 2;                // hdg valid, if calibration value greater equal
```
sarfata: Do you think anyone will ever want to change this? I am not sure we need this.

src/host/config/KBoxConfigParser.cpp
```diff
@@ -46,7 +46,15 @@ void KBoxConfigParser::defaultConfig(KBoxConfig &config) {
   config.serial2Config.inputMode = SerialModeNMEA;
   config.serial2Config.outputMode = SerialModeDisabled;
 
-  config.imuConfig.enabled = true;
+  config.imuConfig.enabled = true;            // enable internal IMU sensor
+  config.imuConfig.enabledHdg = true;         // true if values taken from internal sensor
```
sarfata: I think this should be `headingEnabled` or `enableHeading`. Otherwise it grammatically sounds incorrect.

sarfata: And maybe `enableCompass`? instead of heading?

src/host/config/KBoxConfigParser.cpp
```diff
@@ -46,7 +46,15 @@ void KBoxConfigParser::defaultConfig(KBoxConfig &config) {
...
+  config.imuConfig.enabledHeelPitch = true;   // true if values taken from internal sensor
+  config.imuConfig.calHdg = 2;                // hdg valid, if calibration value greater equal
+  config.imuConfig.calHeelPitch = 2;          // heel (roll) & pitch valid, if calibration value greater equal
+  config.imuConfig.writeN2kHdg = false;       // write heading to N2k bus
```
sarfata: I was thinking we would do this in the NMEA2000 configuration. For example, we would have: 
```
config.nmea2000.attitudePGN = true
```
to turn on generation of this PGN.

This way, no matter how we received it, you can decide if you want to send it to NMEA2000.

I do not know which one is best. Just sharing for discussion. We can start like this and see where it takes us.

src/host/config/KBoxConfigParser.cpp
ronzeiller: of course enableXY OMG – english is not my mother language which you probably saw already :-))

Hmm, we have on some of our boats a Gyro-Compass which delivers Heading and Heel (not sure about pitch)
With enableCompass I think it is not clear because it could be meant the whole IMU thing.
The intention is to stop just hdg values from IMU-sensor.

src/host/config/KBoxConfigParser.cpp
ronzeiller: :-)
E.g.: You had cal 3 for Heading, I took (including) 2 so at least we two have different values.
Same for hdg - I do not want hdg from the BNO055, but heel/pitch will be nice to have.
Especially later, when Leeway is calculated in KBox I will prefer controlled heel-values from KBox

src/host/config/KBoxConfigParser.cpp
ronzeiller: Consider an external DIY-Sensor (my friend is working on this with a NXP-IMU-sensor) which is plugged to serial input. 

I asked myself: "How would it be possible to write HeadingPGN from external sensor only (not from internal IMU), but write attitude from internal sensor?"

But you are right, with settings:
```
config.imuConfig.enableHeelPitch = true;
config.imuConfig.enableHdg = false;
```
it would be enough to to it in the NMEA2000 config.

So I will take out  ```config.imuConfig.writeN2kHdg```

(I do not think that this settings will ever be on a web based config page (because they are too special) but I believe that there should be in the config file.)

ronzeiller:
But anyway I will wait until you merge your config things and then look again and do a clean new PR with the results of this discussion here.

ronzeiller:
closed as covered by new IMUPage [#89](https://github.com/sarfata/kbox-firmware/pull/89)

PR #110, opened 2018-01-02
Write & Read BNO055 Calibration and heel/pitch offsets to/from EEPROM

ronzeiller:
Based on the [PR#89](https://github.com/sarfata/kbox-firmware/pull/89) here are some further developments:

- A second long button press on the IMUMonitorPage within 5 Seconds after the first long press writes offset values for heel & pitch into the EEPROM
- If the BNO055 is fully calibrated (all cals = 3) then the calibration offset values are written to EEPROM automatically every 30 Minutes. 
- At startup of KBox calibration values and offset for heel/pitch are read from EEPROM
- integrated `sysCal` into the` IMUMonitorPage` and `sysCal > 0` into `IMUService::isMagCalibrated()` for better `SKUpdate` of heading. 
May be sysCal should even be 3 as it was in original KBox, just to be sure that heading values are more trustworthy and reliable.

@sarfata :
Please check if PersistentStorage is used ok, especially if the data length of NMEA2000Parameters is calculated right.
And I doubt if it is possible to write int to flash as I do.....
May be it is easier to make you as contributor to change whatever you want?


src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,191 @@
...
   SKUpdateStatic<2> update;
   // Note: We could calculate yaw as the difference between the Magnetic
   // Heading and the Course Over Ground Magnetic.
 
-  /* if orientation == MOUNTED_ON_PORT_HULL */
-  double roll, pitch, heading;
-  roll = eulerAngles.z();
-  pitch = eulerAngles.y();
-  heading = fmod(eulerAngles.x() + 270, 360);
+  if (_config.mounting == VerticalStbHull) {
```
sarfata: We should use a switch statement here. The compiler will throw a warning for the values of the enum that we are not visiting.

src/host/pages/IMUMonitorPage.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include "common/ui/Page.h"
+#include "common/ui/TextLayer.h"
+#include "common/signalk/SKHub.h"
+#include "common/signalk/SKSubscriber.h"
+#include "services/IMUService.h"
+//TODO: damping and automatic damping according to service frequency
+#include "host/config/IMUConfig.h"
+
+class IMUMonitorPage : public Page, public SKSubscriber {
```
sarfata: pushing a commit to fix indentation here.

src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,191 @@
...
+    return true;
+  }
+  ERROR("Error saving IMU Offsets for heel and pitch to flash.");
+  return false;
+}
+
+bool IMUService::readHeelPitchOffset() {
+  struct PersistentStorage::IMUHeelPitchOffsets o;
+
+  if (PersistentStorage::writeImuHeelPitchOffsets(o)) {
```
sarfata: I think you want to call `read` here.

src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,199 @@
...
 
 void IMUService::setup() {
+  _timeSinceLastSetOffset = 0;
+  _offsetRoll  = 0.0;
+  _offsetPitch = 0.0;
+  // Default mounting position: VerticalStbHull
+  uint8_t axisConfig = 0b00001001;
+  uint8_t signConfig = 0b00000000;
+
+  if (_config.mounting == VerticalTopToBow) {
```
sarfata: same here we should use an enum to make sure we cover all the cases.

src/host/services/IMUService.cpp
sarfata: I think this config has a slight issue: The pitch up is measured as negative. It should be positive.

src/host/pages/IMUMonitorPage.h
ronzeiller: Thank You

src/host/services/IMUService.cpp
ronzeiller: > We should use a switch statement here. The compiler will throw a warning for the values of the enum that we are not visiting.
Done

> I think this config has a slight issue: The pitch up is measured as negative. It should be positive

I took your original settings, but I think I mixed up positions. Now port hull is default.
But all positions should be tested, if they are correct, may be you could do it, or I will do, when I am back home next week.

src/host/services/IMUService.cpp
ronzeiller: now switch statement and yet non configured positions commented out in enum

ronzeiller:
@sarfata 
small question about cosmetics:

you code:
```
if ( .... ) {
...
}
else {
...
}
```

In google c++ style guide is different?
https://google.github.io/styleguide/cppguide.html#Conditionals

sarfata:
> @sarfata 
> small question about cosmetics:
> 
> you code:
> ```
> if ( .... ) {
> ...
> }
> else {
> ...
> }
> ```
> 
> In google c++ style guide is different?
> https://google.github.io/styleguide/cppguide.html#Conditionals

I was wrong. Fixing to the right code style.

src/host/services/IMUService.h
```diff
@@ -21,20 +21,49 @@
...
+#include "host/config/IMUConfig.h"
 
 class IMUService : public Task {
   private:
+    IMUConfig &_config;
     SKHub &_skHub;
     Adafruit_BNO055 bno055;
-    uint8_t sysCalib, gyroCalib, accelCalib, magCalib;
+    uint8_t _sysCalib, _gyroCalib, _accelCalib, _magCalib;
+    uint8_t _axisConfig, _signConfig;
```
sarfata: Why did you made these two member variables of the IMUService class? I do not think we need them to be here. They can be local variables when we initialize the sensor like they were before.

src/host/util/PersistentStorage.h
```diff
@@ -65,15 +65,37 @@ class PersistentStorage {
...
+      int offsetHeel;
+      int offsetPitch;
+    };
+
   private:
     // Addresses of elements in flash storage
     static const uint16_t magicAddress = 0;
     static const uint16_t versionAddress = 4;
     static const uint16_t nmea2000ParamsAddress = 8;
+    static const uint16_t bno055CalOffsetsAddress = 20;     // 11 x 2 Byte
```
sarfata: I will replace this by `nmea2000ParamsAddress + sizeof(NMEA2000Params)` this way it will always be correct even if we change the size of the `NMEA2000Params`. Let me know if you see any problem with this.

Of course we will need to upgrade the version too when we do that so that old data is considered invalid.

src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,204 @@
...
+    changed = true;
+  }
+  if (abs(_pitch) < M_PI / 2 ) {
+    _offsetPitch = _pitch * (-1);
+    changed = true;
+  }
+
+  if ( changed ) {
+    INFO("Offset changed: Heel -> %.3f | Pitch -> %.3f", SKRadToDeg(_offsetRoll), SKRadToDeg(_offsetPitch));
+    if (_timeSinceLastSetOffset < 5000) {
```
sarfata: I do not understand why you have `<5000` here. That means you only save if it's been less than 5sec that we started? That means you cannot save if you wait more than 5s after booting?

src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,204 @@
...
+  roll = _roll + _offsetRoll;
+  magCalibration = _magCalib;
+  heading = _heading;
+}
+
+//  Calc offset for making Heel=0 and Pitch=0
+//  (e.g. called with long button press in IMUMonitorPage)
+//  If an setOffset will be done a second time within 5 Seconds,
+//  the values will be stored into EEPROM and loaded at start of KBox
+void IMUService::setHeelPitchOffset() {
```
sarfata: Why do you use `heel` sometimes and `roll` some other times? I think it's a bit confusing and we should probably use only one.

src/host/services/IMUService.cpp
ronzeiller: The first idea was, that with the first "long press" the offset is made temporary and a second "long press" will store it to flash.
Now I think it makes no sense, so the offset could be stored at first press.
But, I think I like the idea, that a second "long press" could reset offset to zero.

src/host/services/IMUService.cpp
ronzeiller: The problem is that e.g. openCPN or any display shows "heel" and not "roll"
And also when we come to performance calculations everywhere is mentioned "heel"

Remember, when I asked you to change the wording to "heel"?
It is a pity, that SignalK has chosen "roll"....

So I took "heel" as much as possible and let roll where it has to be (IMU sensor, SignalK)

src/host/services/IMUService.cpp
```diff
@@ -26,38 +26,204 @@
...
+  }
+  // update NavigationAttitude if quality (= calibration value) is ok
+  if (isHeelAndPitchCalibrated()) {
+    update.setNavigationAttitude(SKTypeAttitude(/* roll */ _roll + _offsetRoll, /* pitch */ _pitch + _offsetPitch, /* yaw */ SKDoubleNAN));
+    //DEBUG("Heel = %.3f | Pitch = %.3f | Heading = %.3f", SKRadToDeg( _roll + _offsetRoll), SKRadToDeg( _pitch + _offsetPitch), SKRadToDeg(_heading));
+  }
+  _skHub.publish(update);
+
+  // Save calibration values to EEPROM every 30 Minutes, if BNO055 is fully calibrated
+  if (bno055.isFullyCalibrated() && _timeSinceLastCalSave > 1800000) {
```
sarfata: `_timeSinceLastCalSave` is never initialized. You should initialize it to the value of the timer so that it automatically saves the first time it's fully calibrated. And this should be a `static const` member of the class.

src/host/util/PersistentStorage.h
ronzeiller: Aah, now I know why you did the version!

The disadvantage could be:
Other values on higher addresses stored, e.g. Logging of total NM will get lost too
And you will have more writing-processes at the end.

May be there should be space reserved for permanent values or clear values which have fixed size and you put things which are not clear yet to higher addresses?

src/host/services/IMUService.cpp
ronzeiller: First point is clear.

But why static const?
If the sensor looses calibration and later on will get calibrated again, the new values should be written again to flash I think.

src/host/services/IMUService.h
ronzeiller: Because I was not aware, that it makes a lot of difference.
Sure I will bring them back!

sarfata: It just makes the object a little bigger in memory. These classes are allocated on the heap when we start the program. The local variables inside functions are allocated on the stack. If possible, it's better to use stack than heap for us.

src/host/util/PersistentStorage.h
sarfata: yeah right now it does not handle that well.

src/host/services/IMUService.cpp
sarfata: This is also what NMEA2000 is using: https://github.com/ttlappalainen/NMEA2000/blob/master/src/N2kMessages.h#L403

It's ok to use heel on the display but I do not think we should use it in the code. Having two names for the same thing is just confusing and I am tempted to follow what everyone else is doing here.

sarfata:
No problem with your change on calibration. Sounds good to me.

I think this is all as good as it can get for now so merging now and we can still do more changes later! Thanks a lot. Major contribution here!

⛵️

sarfata:
Ah we forgot to update the README! I will push another commit summarizing the changes. Please help me remember in the future ;)

PR #113, opened 2018-01-06
Rz performance extension for discussion

ronzeiller:
**Rough Idea for performance calculations.
Beginning with correction of boat speed by calibration values and correction of measuring failure due to leeway. Leeway calculation by raw boat speed and heel.**

This corrected boat speed should also be written to N2k-bus to have it on boat's MFD.

The idea is that raw values (mainly boat speed and apparent wind) coming from N2k-bus should be corrected before SKUpdate. (switched by config settings of course)

As everything depends on boat speed this will be the trigger for all following calculations of
- true wind (config to choose if a user want to take N2k True Wind or calculate by KBox)
- current
- performance (target boat speed, % performance, etc.)

For wind based calculations it could be necessary to have an array to store the last n values for calculate a mean value (e.g. by moving average), but for the beginning the last actual apparent wind values would be taken.

An example of a polar data file (Expedition style) and an idea for a boat speed correction table are enclosed in extras/performance
  
**Terrible!**
I did merge by accident and now even after revert all this files are still shown, so no one can see the performance updates. 👎 
What to do now?
Should I make a new branch with this new files?


sarfata:
> Terrible!
> I did merge by accident and now even after revert all this files are still shown, so no one can see the performance updates. 

You could have changed the base branch so that we would have looked at the differences between the imu branch and this one. Otherwise yes you can create a new branch and just cherry-pick the commit you want in it.

Another way to do it is to use `git rebase -i` in the commend line to get rid of the comments you do not want and only keep what you want.

It's good to have multiple solutions with git because the first one you try does not always work ;)

PR #115, opened 2018-01-08
Send all NMEA2000 messages through SKHub

sarfata:
- Will enable NMEA2000 to appear in SK data
 - Will also enable it to appear in NMEA output

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,11 +192,14 @@ been possible!**
 
 ## Changelog
 
+ * 2018 01 07
+   * Data coming from NMEA2000 bus is now sent to SignalK websocket output
+   and will be converted to NMEA0183 (if a conversion is available).
  * 2018 01 04
    * Add support for parsing PGN127257 Attitude (roll/pitch/yaw)
    * Also improved the output in NMEA and SignalK format when some attitude
    values are unknown
-  * 2017 12 28
+ * 2017 12 28
    * KBox now supports a configuration file (`kbox-config.json`) on the sdcard.
    * KBox supports sending data to NMEA outputs
    * Generated NMEA sentences can be configured independently for each output
--- a/src/host/services/NMEA2000Service.cpp
+++ b/src/host/services/NMEA2000Service.cpp
@@ -23,12 +23,14 @@
 */
 
 #include "NMEA2000Service.h"
+
 #include <KBoxLogging.h>
 #include <KBoxHardware.h>
 #include <TimeLib.h>
 #include "common/stats/KBoxMetrics.h"
 #include "common/algo/crc.h"
 #include "common/version/KBoxVersion.h"
+#include "common/signalk/SKNMEA2000Parser.h"
 #include "host/util/PersistentStorage.h"
 
 static NMEA2000Service *handlerContext;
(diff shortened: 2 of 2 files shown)


ronzeiller:
@sarfata 
does this mean, that due to additional
```
SKNMEA2000Parser p;
const SKUpdate &update = p.parse(SKSourceInputNMEA2000, msg, SKTime(0));
```
all PGN's get parsed twice now?

sarfata:
@ronzeiller I looked around to confirm before replying but I do not think we are parsing NMEA2000 messages anywhere else. This was missing. Where else do you see parsing of NMEA2000 messages?

The only other place where it's used is in tests and in `sktool` but I think it had not been added to KBox runtime yet.

sarfata:
I think we could have filter at parsing in the future too yes. Nobody has really requested it yet and I am trying to limit myself to things that people have a real use for but in the future, why not.

ronzeiller:
> I think we could have filter at parsing in the future too yes. Nobody has really requested it yet and I am trying to limit myself to things that people have a real use for but in the future, why not.

At the moment we are not parsing that much PGN's so this is definitely something for later...

> @ronzeiller I looked around to confirm before replying but I do not think we are parsing NMEA2000 messages anywhere else. This was missing. Where else do you see parsing of NMEA2000 messages?

Uups, I mixed that up with former KBox version, where the parsing was by the visitor.

Is there anywhere a button for deleting stupid questions? :joy:

ronzeiller:
But one thing turned up:
When reading and parsing NMEA2000 PGN's **together** with WiFi Service, then switching to the next page is not possible anymore. (KBox does not react most of the times on button pressed or turned).
If I switch of PGN-generation in Timo's simulator, then suddenly page setting is possible again.
On the other hand, if I deactivate WiFi by config setting then everything is normal (PGN's get parsed, switching pages possible)

If the PGNs are reduced in the simulator it seems that it will behave better.

On the other hand I have periodically following messages in the serial monitor, where I have no idea:

```
E WiFiService.cpp:63 Invalid command received from WiFi module (id=80 size=12)
D WiFiService.cpp:65 > P
WI KommandHandler.cpp:32 Received error frame!

```

There is no hurry. If you find the time, could you please have a look if you experience the same, or if it is just with my (pseudo-)KBox here?

sarfata:
How many PGNs are you generating with Timo's module?

How did you connect the wifi module to the teensy? It's running at 115200 bps for now but we could accelerate it (and maybe increase buffer too) if we are loosing data. I have not seen it yet, I should probably try the simulator to increase the traffic and see this type of issue.

ronzeiller:
At first I activated all PGNs including rudder and speed :-)
The reduced PGNs were full Wind/temp and Navigation without the fast (0,1) only.

The WiFi module  is connected to Serial1 (Pins 0,1)
It is programmed by sending `[env:esp] ` directly to the MCU, but without the `extra_scripts = tools/platformio_cfg_esp.py`

ronzeiller:
nota bene, may be it helps to find out what happens:
I deactivated the block
```
// Now send in JSON format
StaticJsonBuffer<1024> jsonBuffer;
SKJSONVisitor jsonVisitor(jsonBuffer);
.....
```
and activated all PGNs in simulator again and page switching is back again without delay.

PR #117, opened 2018-01-08
Introducing Performance Calculations

ronzeiller:
**Performance calculations.
Beginning with correction of boat speed by calibration values and correction of measuring failure due to leeway. Leeway calculation by raw boat speed and heel.**

~~The Performance class is always called between incoming datas  from a sensor (mainly boat speed and apparent wind) and before this uncorrected datas would be written to SKUpdate. 
This could be in NMEA2000Parser as well as in NMEAParser as there is first analyzed which datas are coming in.~~

Update on 9.2.2018:
The Performance class is a service (Task) started in main.cpp
  When there is an update for heel coming, the value will be stored.
  When there is an update for boat speed through water coming we calculate
  first leeway and second correct boat speed with cosine of leeway angle.

As all performance calculations depend on boat speed 1.) this will be the trigger for all following calculations of
- true wind 2.)
- current
- performance (target boat speed, % performance, etc.)

An example of a polar data file (Expedition style) and an idea for a boat speed correction table are enclosed in extras/performance.

**Footnotes:**
1.) boat speed is coming in just 1Hz, all other values needed for performance calculations are coming at least 4 times more often.

2.) For wind based calculations it could be necessary to have an array to store the last n values for calculate a mean value (e.g. by moving average), but for the beginning the last actual apparent wind values would be taken.

**Sources to show where to the journey could go:**
1. [Yacht Performance analysis with computers 1981 by D. Pedrick and Richard McCurdy](http://vm2330.sgvps.net/~syrftest/images/library/20160217160905.pdf)
2. [How Ockam calculates things](http://www.ockam.com/docs/Calculations.pdf)
3. [ORC Speed Guide Explanation](http://www.orc.org/rules/Speed%20Guide%20Explanation.pdf)


extras/config/default-config.json
```diff
@@ -50,5 +50,12 @@
   "nmea2000": {
     "txEnabled": true,
     "rxEnabled": true
+  },
+  "performance": {
+    "enabled": true,
+    "boatSpeedCorrectionTable": "boatspeedCorr.cal"
```
sarfata: missing a comma at the end of the line. this will not load properly.

src/host/config/PerformanceConfig.h
```diff
@@ -0,0 +1,40 @@
...
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+*/
+
+#pragma once
+#include <WString.h>
+
+struct PerformanceConfig {
+  bool enabled;
+  String boatSpeedCorrectionTable;
```
sarfata: this is just the filename right? Maybe we should include `filename` in the name of the variable so it's obvious.

src/host/config/PerformanceConfig.h
```diff
@@ -0,0 +1,40 @@
...
+  int hullFactor;         // usually between 8....10, here multiplied by 10
+  int windSensorHeight;   // in cm
+  String polarDataFile;
```
sarfata: `polarDataFilename`

When I first read this, I thought we were loading the entire file in the string here.

src/host/performance/Performance.h
```diff
@@ -0,0 +1,48 @@
...
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+*/
+/*
+  Annotation:
+  As we are here for sailboat performance calculations we use the common used
+  "heel" instead of "roll" as in the rest of KBox code.
```
sarfata: lol

src/host/config/PerformanceConfig.h
sarfata: Or maybe we should actually read these data files when we load the config and store the in-memory representation after parsing. Have you worked on it already? Parsing those polar files is not going to be trivial.

ronzeiller: Yes, the polarfile parsing will be an own adventure...
I did not start to work on that, just put it into the code actually not to start with it at the beginning, but more to give you an oversight of things we will have to bear in mind.

Some ideas for the polar data reading/calculating you will find at Thomas Rauch's work for openCPN: https://github.com/tom-r/tactics_pi/blob/master/src/performance.h

I think we should open an own issue for the whole polar data file thing to discuss it there.

src/host/performance/Performance.h
ronzeiller: I was sure that you will like that :-)))

extras/config/default-config.json
ronzeiller: done

src/host/config/PerformanceConfig.h
ronzeiller: good idea!
Done

ronzeiller:
As mentioned in the first comment of the PR I would suggest that we start with correction of boat speed which is essential.
Second could be own, trusted True Wind calculations which we can write to several outputs.
So we are at least able to connect KBox to openCPN with tactic_pi Plugin and have the value on the MFD for the beginning.

I really hope you like this ideas, because for extending the internal structure of KBox code for config and integration of this performance calcs I need your help to get it started.

sarfata:
> I really hope you like this ideas, because for extending the internal structure of KBox code for config and integration of this performance calcs I need your help to get it started.

Yes I like! I will try to help but I already owe you multiple things and have a few PR in flight. You are putting a lot of pressure on me and I love it!

ronzeiller:
> > I really hope you like this ideas, because for extending the internal structure of KBox code for config and integration of this performance calcs I need your help to get it started.

`Yes I like! I will try to help but I already owe you multiple things and have a few PR in flight. You are putting a lot of pressure on me and I love it!`

👍  

Take your time!
Don't feel pushed!

But, beginning of April is my regatta and I would like to have lot of things ready then :grin:

ronzeiller:
Next TODOs:
  1. extend config to choose from where heel is taken (N2k Bus, internal sensor
      or NMEA0183 sensor on serial input)
  2. read correction file for boatspeed offsets for upright and heeled sailing
  3. extend class for corrections on Apparent Wind and calculations of True Wind
4. read the polar data file
5. do performance calculations

ronzeiller:
@sarfata 
Hi Thomas,
performance calculations are improving. 
Next step is True Wind calculations and reading the polar data file.

But I would need your help please, because  just to `#include <SDFat>` or  `#include <KBoxHardware.h>` for reading file on SD-Card breaks the test and I do not know why.

sarfata:
@ronzeiller Sorry I have not been very present lately. Lots of work on my boat which has priority at the moment.

It is very hard to make the code that accesses the hardware testable and the class KBoxHardware.h is not testable at the moment. What I recommend is splitting your code so one part is independent of the hardware, does not include KBoxHardware.h and you can test this one. Then you can have another class that connects it to the hardware and this one you will not test.

For example: make `Polar.h` take a String and read the data from the String. This way it does not need access to the SDCard and you will be able to write tests for it. Then somewhere else (maybe in your service) you load the content of the file into a string and pass it to `Polar.h`.

ronzeiller:
@sarfata 

> @ronzeiller Sorry I have not been very present lately. Lots of work on my boat which has priority at the moment.

No problem, I think that there are lot´s of other things out there which might have more priority :-)


> For example: make Polar.h take a String and read the data from the String. This way it does not need access to the SDCard and you will be able to write tests for it. Then somewhere else (maybe in your service) you load the content of the file into a string and pass it to Polar.h.

I tried to read from SD-Card in my service, but 
`make test`
fails completely as soon as 
`#include <SdFat.h>`
is added.

Error:
```
In file included from .piolibdeps/SdFat/src/BlockDriver.h:.piolibdeps/SdFat/src/SysCall.h:37.piolibdeps/SdFat/src/SysCall.h:37:2: error: "Unknown system"
```

It seems that I will have to read the Polardata file in `main.cpp` after 
```
if (KBox.getSdFat().exists(polarDataFileName)) {
....
}
```
and pass the string to the PerformanceService from there.
This is working without breaking test.

ronzeiller:
The performance calculations are now called by the SKHub.
Performance class is completely new written.

You will find new code (where I try to stay as close as possible to the whole KBox project, also trying to merge all future improvements from the original KBox firmware) here in the several Sailmax-CU branches:

https://github.com/ronzeiller/kbox-firmware/

(Part of this discussion here will be copied there too.)

PR #118, opened 2018-01-09
Feature/50 hall speed sensor

sarfata:
Allows a user to connect a good-old paddle wheel to KBox and get a 10Hz speed sensor that speaks NMEA and NMEA2000. 

Work in progress. @ronzeiller you might be interested in this too.

### DIFF
--- a/lib/KBoxHardware/src/board.h
+++ b/lib/KBoxHardware/src/board.h
@@ -98,3 +98,6 @@ static const pin_t supply_analog = A14;
 
 /* SDCard Interface */
 static const pin_t sdcard_cs = 15;
+
+/* Analog Sensor Interface */
+static const pin_t paddle_wheel_input = 25;
--- a/src/common/signalk/SKNMEA2000Converter.cpp
+++ b/src/common/signalk/SKNMEA2000Converter.cpp
@@ -56,7 +56,9 @@ void SKNMEA2000Converter::convert(const SKUpdate& update, SKNMEA2000Output& out)
 
   // PGN 128259: Boat Speed
   if (update.hasNavigationSpeedThroughWater()) {
-
+    tN2kMsg msg;
+    SetN2kBoatSpeed(msg, 0, update.getNavigationSpeedThroughWater(), N2kDoubleNA, N2kSWRT_Paddle_wheel);
+    out.write(msg);
   }
 
   // PGN 129026: Fast COG/SOG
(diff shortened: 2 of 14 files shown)


ronzeiller:
@sarfata 
Yes I am interested, in deed.

Just a few thoughts:
- I would recommend to name Services less generic, as they are very special for a dedicated design.
- Another question is, how to connect the sensor?
- Wouldn't it make sense to build an own box for translation of ticks and conversion to N2k or NMEA0183 and send the signal over already existing cables, (similar to this: https://www.ensslin.com/airmar-dx900/a-20724/), or take the 12V from somewhere and send the signal over Bluetooth?
- Nice to have: the correction table mechanism (reading, interpolationg...) needed to correct for non linearity of the ST650 ticks could nicely also be used for correction of common speed transducers.
- Lot of work to get a rather old sensor calibrated, converted, filtered,...... just for a hand full users?

sarfata:
Yes you need a separate cable and probably I will make a small PCB with an optocoupler for this use-case. Why make a new product when KBox can do it? It's easier to have only one design. With config, you could have a KBox doing just this conversion in the bow. It's still much cheaper than any of the commercial options.

This is very interesting to people who have an old sensor and want to send the data on NMEA2000 bus. I had that on my boat and hauling the boat to drill a new hole of a different size for the new sensor was not a simple operation. 

The other *very* interesting thing for racers is that we can send speed at 10Hz instead of the 1Hz of the DST800. So it is much cheaper, and actually much better too.

I agree that other work on calibration/correction should apply to any sensor. 

How would you name it?

ronzeiller:
> Yes you need a separate cable and probably I will make a small PCB with an optocoupler for this use-case. Why make a new product when KBox can do it? It's easier to have only one design. With config, you could have a KBox doing just this conversion in the bow. It's still much cheaper than any of the commercial options.

I agree, that this is a great solution.
And this is exactly what I meant, that the KBox Software could easily manage such sensors.
(It´s only a small step from a PCB with optocoupler to a PCB with a readymade Teensy.....,
may be with additional Bluetooth function in the future - thinking also of external, solar powered displays, like Tacktick/Raymarine has made them.....)

> The other very interesting thing for racers is that we can send speed at 10Hz instead of the 1Hz of the DST800. So it is much cheaper, and actually much better too.

💯 


> How would you name it?

Something like `PaddleWheelService`, or `ST650Service` (depending of how many other similar paddlewheel sensors are out there), `BSTransducerService` for Boat Speed Transducer....?

BTW: I plan to go for an external IMU-Sensor based on the NXP, which could also be directly connected to KBox (I2C)....

PR #119, opened 2018-01-10
SDCard and DataFormat config

ronzeiller:
Aim will be:

* set different possibilities for logging datas, like
    * PCDIN from N2k only
    * mixed mode NMEA + PCDIN
    * NMEA only
    * w/o timestamp
* to configure KBox as pure logging device
* add debug messages to log file
* create new log file for each day
* add file attributes date + time to log files
  
Other todo's:
Correct empty line in log file

At the moment config settings are bound to `Services`, but we will need them on other places in the code too. (e.g.: `SDCardConfig` in `KMessageNMEAVisitor` to prevent writing `SeaSmart` if not set in config, but how this could be done?)


src/common/signalk/KMessageNMEAVisitor.cpp
```diff
@@ -35,10 +35,10 @@ void KMessageNMEAVisitor::visit(const NMEA2000Message &n2km) {
 
   const tN2kMsg& msg = n2km.getN2kMsg();
 
+  // TODO: integrate SDCardConfig to check if SeaSmart is enabled
```
sarfata: You could just pass a boolean to this visitor constructor. When we create it we know if we need to generate PCDIN or not.

sarfata:
> At the moment config settings are bound to Services, but we will need them on other places in the code too. (e.g.: SDCardConfig in KMessageNMEAVisitor to prevent writing SeaSmart if not set in config, but how this could be done?)

Look at `SKNMEAConverterConfig`. It's already an example of config that is not tied to a service and that is already re-used in multiple places.

ronzeiller:
> > At the moment config settings are bound to Services, but we will need them on other places in the code too. (e.g.: SDCardConfig in KMessageNMEAVisitor to prevent writing SeaSmart if not set in config, but how this could be done?)

> Look at SKNMEAConverterConfig. It's already an example of config that is not tied to a service and that is already re-used in multiple places.

Thank You, I´ll have a closer look at this

src/common/signalk/KMessageNMEAVisitor.cpp
ronzeiller: I did not give up the idea, that KBox could also write Actisense Binary to Serial output or WiFi....
It´s something for future works, if we get bored and have nothing else to do :-)

But for now a boolean should make it perfectly!

sarfata:
Oh yes that is a good idea. Pass an enum instead of a bool. Works too!

ronzeiller:
@sarfata 
**Concept for different dataFormatTypes for In/Output added:**
Dataformats are e.g.: 
- NMEA0183 *)
- Seasmart ($PCDIN)
- NMEA_Seasmart Combination of both **)
- SignalK ?

**future possible formats:**
- LoggingDataAsCSV (e.g. Expedition style)
- ActisenseBinary
- ActisenseText (logging alternative to Seasmart)
- CAN-Boat style N2k
- Gofree
- Seatalk

Could also be used for serial in/out -> may be make serial `config.enable` (as other Services have) and cancel enum with `SerialModeNMEA` ?

*) including proprietary sentences, were openCPN could be configured to understand them
**) Seasmart is not very common here, and as a user I want to turn it off, but it is a compact format for logging

ronzeiller:
Could you please have a look @sarfata?
More for discussion than to merge, with lot's of DEBUG messages.....

Now it is possible to set different DataFormats for SD-Card logging, WiFi and the plan is that it should work with Serial1/2 too.

The question is, if the way I am doing the `DataFormatConfig.h` in `KBoxConfigParser` is the right thing to do?

But (also in master branch) there are some issues:
- on SD-Card there are only PCDIN/Seasmart stored but never NMEA
- internal sensor datas are not going to be logged SD-Card

Issues with my PR:
- NMEA is always sent to WiFi -> where can I integrate my config to stop it?
- Input/Output on Serial is not tested at all, I have just done NMEA2000 and internal sensors so far.

ronzeiller:
No idea why make validation failed....

```
src/sktool/sktool-test.py --implementation sktool src/sktool/tests/*json
/bin/sh: src/sktool/sktool-test.py: Permission denied
make: *** [validation] Error 126
```

ronzeiller:
outdated by https://github.com/sarfata/kbox-firmware/pull/134

sarfata:
It failed because the permission of the files changed in one of your commit. The file is not executable anymore on UNIX systems. This commit was the problem: https://github.com/sarfata/kbox-firmware/pull/119/commits/c1d6dbb4de8739f85f799086a10cfc0be80dad5b

PR #120, opened 2018-01-10
some nitpicks

ronzeiller:


src/common/signalk/SKNMEAParser.cpp
```diff
@@ -133,10 +133,8 @@ const SKUpdate& SKNMEAParser::parseMWV(const SKSourceInput& input, NMEASentenceR
 
   // Wind sensors return a number between 0 and 360 but we want an
   // angle in radian with negative values when wind coming from port
-  if (windAngle > 180) {
-    windAngle = windAngle - 360;
-  }
   windAngle = SKDegToRad(windAngle);
+  SKNormalizeAngle(windAngle);
```
sarfata: I think you meant `windAngle = SKNormalizeAngle(windAngle);`

ronzeiller: Yesss :-)

I was finding it incidentally and thought it should be the same as NMEA2000, so 180° is already "from port"

sarfata:
I fixed a very minor test issue and added a test for the case where wind angle is >180.

⛵️

PR #124, opened 2018-01-11
Adafruit bno055 update + added position configs

ronzeiller:
Update KBox-BNO055 library to Adafruit BNO055 Version 1.1.6
https://github.com/adafruit/Adafruit_BNO055

Corrected a bug in cal-offsets.
Hopefully heading will be better now with true calibration offset values.....


src/host/util/PersistentStorage.h
```diff
@@ -67,7 +67,7 @@ class PersistentStorage {
 
     struct IMUCalibration {
       uint8_t mountingPosition;
-      uint8_t calibrationData[22]; // NUM_BNO055_OFFSET_REGISTERS
+      int calibrationData[22]; // NUM_BNO055_OFFSET_REGISTERS
```
sarfata: why int? it's an `int16_t`, no?

That is a huge bug - and it was all me! Definitely explains some of the weird results we were seeing. Thanks a lot for finding that - and sorry for screwing up your code ;)

sarfata:
Did you try this? It does not compile? Looking now.

src/host/util/PersistentStorage.h
sarfata: Actually, there is no bug.

We are using the function `getSensorOffsets(uint8_t* calibData)` which reads the register as an array and writes them back as an array so the changes Adafruit did will not impact us. And we are reading the correct size: 22 registers of 8bits each.

So I do not think this PR will improve anything unfortunately.

We can merge the Adafruit update to stay current but that will not change anything for us.

ronzeiller:
Oh, so sorry!!!
It was a late night short thought to change calibrationData[22] too.
Now we are back at Adafruits update. (With our own extensions)

Of course you are right, that it doesn't make any improvement for us :-(

BTW, I discussed there the matter of using the sensor in vertical mode:
https://github.com/adafruit/Adafruit_BNO055/pull/38#issuecomment-356934103

> ......remember to apply offsets to your accelerometer accordingly (page 31) so that gravity is removed from Y instead of Z .....

May be this is interesting for vertical positioning?

ronzeiller:
@sarfata 
Now the PR consists of:
- some (not needed) parts of Adafruit BNO055 library updated to 1.1.6
- config + IMUService now for all horizontal mounting positions
- Bosch placement comments for positions added
- enum names changed to be equal with config names

ronzeiller:
@sarfata
Some more improvements and add ons:
PR actually consists of:

* some (not needed) parts of Adafruit BNO055 library updated to 1.1.6
* added and renamed mounting positions. Now they are (hopefully) easy to understand and equal for KBox and a free floating BNO055 sensor connected to KBox.
* enum names equal with config names to make config easier for a user
* checked all mounting positions for correct output
* added kbox-config.json to .gitignore, so personal config might stay in extras/config and can easily be changed/uploaded by command:
`tools/kbox.py --port /dev/nameOfUSB fwrite extras/config/kbox-config.json kbox-config.json`


(Bosch placement comments for positions deleted, as they are not helpful)

ronzeiller:
Updated PR to latest commits in master
Update: now the color of the displayed values correspond to if they are sent as SKUpdate to SKHub.
(e.g. if hdg shows in red it shows that there are not sent any NMEA0183 values to serial output)

platformio.ini
```diff
@@ -31,7 +31,7 @@ build_flags =
     -DSERIAL3_RX_BUFFER_SIZE=256 -DSERIAL3_TX_BUFFER_SIZE=256
 platform = teensy
 framework = arduino
-board = teensy31
+board = teensy36
```
sarfata: I cannot merge with this in it.

Why don't you create a section `[env:host36]` so this problem does not come up?

src/host/config/IMUConfig.h
```diff
@@ -27,22 +27,37 @@
...
 enum IMUMounting {
-  VerticalStbHull,
-  VerticalPortHull,
-  VerticalTopToBow,
-  //VerticalTopToStern,
-  HorizontalLeftSideToBow,
-  //HorizontalRightSideToBow
+  verticalPortHull,           // KBox mounted on port hull
+  verticalStbHull,            // KBox mounted on stb. hull
+  verticalRightSideToBow,     // KBox mounted on port hull
```
sarfata: Why have two options that are the same thing? It looks like `verticalPortHull == verticalRightSideToBow`? Or maybe I am missing something?

sarfata (changes requested):

platformio.ini
ronzeiller: Sorry, I did not change back after testing, but the `[env:host36]` is so easy and cute!
Thanks for this!

src/host/config/IMUConfig.h
ronzeiller: Yes if I remember right, it is the same. verticalPortHull is the "deprecated" naming convention which I believe you have in your config. If you switch to the "more detailed naming convention" then we do not need it anymore.

If you like it in principle and if it is easy for you to change, please do so. 
Otherwise I think the right way is to merge your new commits, add the `[env:host36]` and delete the `verticalPortHull`, `verticalStbHull` settings in this PR, right?

sarfata:
Welcome back ;) Let me take a look tomorrow and get back to you! Glad to hear from you!!

platformio.ini
```diff
@@ -68,6 +68,21 @@ extra_scripts = tools/platformio_cfg_gitversion.py
...
+    -DSERIAL1_RX_BUFFER_SIZE=512 -DSERIAL1_TX_BUFFER_SIZE=512
+    -DSERIAL2_RX_BUFFER_SIZE=256 -DSERIAL2_TX_BUFFER_SIZE=256
+    -DSERIAL3_RX_BUFFER_SIZE=256 -DSERIAL3_TX_BUFFER_SIZE=256
+platform = teensy
+framework = arduino
+board = teensy36
+lib_deps =
+  ${common.lib_deps_common}
+  ${common.lib_deps_teensy}
+extra_scripts = tools/platformio_cfg_gitversion.py
```
sarfata: I think you may need to add `lib_ignore = ${common.incompatible_libs_teensy}` when you merge. Not sure ... Please give it a try. I have added it for `env:host`.

src/host/config/IMUConfig.h
sarfata: I like the new naming and I think it's good to remove the old one to keep the more explicit one. 👍

src/host/services/IMUService.cpp
```diff
@@ -42,15 +57,26 @@ void IMUService::setup() {
   // Where each can be 00:X, 01: Y, 10:Z
   // Sign is the 3 lsb: 00000xyz
   switch (_config.mounting) {
-    case HorizontalLeftSideToBow:
+    case horizontalTopToBow:
```
sarfata: Do you have time to test and make sure it works well?

If not, can you move all the ones that are untested at the bottom and add a comment to say that they have not been tested yet?

src/host/services/IMUService.cpp
```diff
@@ -71,29 +97,36 @@ void IMUService::loop() {
...
+    case verticalBottomToBow:
+    case horizontalBottomToBow:
       _roll = SKDegToRad(eulerAngles.y());
       _pitch = SKDegToRad(eulerAngles.z());
       _heading = SKDegToRad(fmod(eulerAngles.x() + 180, 360));
     break;
-    case HorizontalLeftSideToBow:
-      _roll = SKDegToRad(eulerAngles.y());
-      _pitch = SKDegToRad(eulerAngles.z()) * (-1);
+    case verticalTopToBow:
```
sarfata: Same here. Can you let me know which one you have tested and which ones are untested?

sarfata (commented):
Looks good. I agree we can merge this quickly. I would just like to clarify which configurations have been tested so we know and we can document what works and what does not (yet).

platformio.ini
ronzeiller: I am afraid I did some mistake on the last merge of your updates, because
`[env:host-teensy36] [env:program-esp]  [env:program-esp-teensy36]`
are already updated in your uptodate master branch.
It was doubled in the PR by mistake.

did a
`git checkout upstream/master platformio.ini`
now

src/host/services/IMUService.cpp
ronzeiller: @sarfata good that you asked for more tests ;-)
Found and corrected a bug in inverting angles

Tested all 8 position variants successfully now

(You could do a final test with one of the positions with KBox, just to be sure I got it right. 
Not 100% sure with the BNO chip positioning in KBox)


Just wondering why it doesn´t do any new checks....

PR #125, opened 2018-01-24
NMEA2000ParserConfig discussion

ronzeiller:
@sarfata 
As a user I want to en/disable single PGN's for parsing.

At the moment we do not have all PGN's parsed, but soon there will be more, which makes it interesting to switch off unwanted PGN's.

Also it is possible to switch off those PGN's coming from bus, which are "produced" by built-in sensors. (e.g. Heading, Heel/Pitch)

I started with the config of 3 PGN's as a proof of concept.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ CMakeListsPrivate.txt
 
 # CLion
 cmake-build-debug
+
+# individual user's config file
+extras/config/kbox-config.json
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,9 @@ logs:
 logs-n2kanalyzer:
 	miniterm.py --rts 0 $(PORT) 115200 | grep 'TX: $$PCDIN' |perl -pe 's/.*TX: //' |analyzer -json
 
+reboot:
+	tools/kbox.py --port $(PORT) reboot
+
 mfg:
 	platformio run -e mfg
 
--- a/README.md
+++ b/README.md
@@ -192,6 +192,22 @@ been possible!**
 
 ## Changelog
 
+ * 2018 02 16 - v1.2.2
+   * Go back to older version of ESP8266 framework (2.3 via framework 1.5) to
+     get rid of random crashes. Still need to investigate more but we do not
+     need 2.4 at the moment.
+ * 2018 02 15 - v1.2.1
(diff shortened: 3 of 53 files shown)


ronzeiller:
Just asking myself if `SKNMEA2000ParserConfig` should be more generic, or should be integrated to `SKNMEA2000Config`

For example:
If a user (do not) want(s) to write Heading and Attitude from built-in IMU-sensor to the N2k-bus.
The `SKNMEA2000Config` could be connected to the `IMUConfig`.

Then in `SKNMEA2000Converter` it could be checked:
```
if (txEnabled == true &&
an update is coming from IMUService && 
config.imuConfig.nmea2000.attitude_127257_enabled == true) {
    out.write(msg);
}
```

Same for `SKNMEAConverterConfig`:
A more generic `SKNMEA0183Config` could be used not only for output but for input too, and could be extended to chose different Talker-IDs!
 
E.g.: to write `KB` as Talker-ID for NMEA0183 sentences from KBox internal sensors.

The problem at the moment is: 
if Hdg is coming from N2k AND from IMU-sensor the output NMEA0183 sentence have the same Talker-ID's but different datas!
If the Talker-ID would be different, then the correct sentence could be selected in the navigation program.

ronzeiller:
The problem is more complicated as I thought at the beginning.
Especially if we calculate (and send back to the bus) datas which are based on incoming datas. E.g.: Boat Speed (BSP) or we produce own datas like Heading (HDG) by internal sensor.
You might be able to choose the "right" PGN on the MFD by it´s source, but as far as I saw not in navigation program as Expedition. So the value "jumps" between those different values.

What will be needed from my point of view:
1. en/disable incoming PGNs 
if a PGN is not needed internal, nor for navigation program or logging it should be dropped.
2. en/disable outgoing (written back to the bus) PGNs by it´s source
BSP from the bus is needed for calculations, but own BSP should be sent to nav program and logging

Similar task for NMEA0183. It is definitely nice to have own TalkerID, so at least with openCPN it is possible to choose "the right" sentence. But not all nav programs are capable to do so.

At the moment we can dis/enable sentences by "HDM", "RSA", etc. for Serial in and Serial out, but we can not choose to drop incoming HDM and send HDM from internal sensor.

